PR #10, opened 2020-11-07
remove pooling for now

fmoor:
Remove pooling until there is clear plan for how it should work. See https://github.com/edgedb/edgedb-go/pull/1#discussion_r517530391 for discussion.


fmoor:
I agree that there should be pooling built in, but the current pooling was just a place holder and doesn't work the way I think would be best. In some ways it doesn't work at all i.e. clients never disconnect from the server. I would like to nail down the details and start from scratch.

fmoor:
Oops didn't mean to merge...

PR #17, opened 2020-11-12
add buffer abstraction

fmoor:
This change adds a buffer abstraction that panics if the buffer is over or under read.

image


1st1:
Can you run benchmarks before and after this change?

protocol/buff/reader.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// PopUint64 reads a uint64 and advances the buffer.
+func (b *Message) PopUint64() uint64 {
+	if len(b.bts) < 8 {
+		panic("buffer overread")
+	}
+
+	val := binary.BigEndian.Uint64(b.bts)
+	b.bts = b.bts[8:]
```
1st1, line 107: I'm worried about two things:

1. The `Message` struct -- we need to spend resources on allocating those structs.
2. Using slices to advance buffers -- Go runtime will probably create a new sliced buffer object for every slice operation, right?

protocol/codecs/base_scalar.go
```diff
@@ -137,18 +137,15 @@ func (c *UUID) Type() reflect.Type {
 }
 
 // Decode a UUID.
-func (c *UUID) Decode(bts *[]byte, out reflect.Value) {
-	protocol.PopUint32(bts) // data length
+func (c *UUID) Decode(msg *buff.Message, out reflect.Value) {
 	p := (*types.UUID)(unsafe.Pointer(out.UnsafeAddr()))
-	copy((*p)[:], (*bts)[:16])
-	*bts = (*bts)[16:]
+	copy((*p)[:], msg.PopBytes())
```
1st1, line 142: I'm just curious: why do we do unsafe pointer stuff here and don't do it in `(c *Bytes) Decode()`?

1st1 (approved):
This is really great work in the right direction. I worry that the read buffer is a bit more high level than it should be.

Our implementation in https://github.com/MagicStack/py-pgproto/blob/2dc7cc4635b08418dfbc6105b45947244102e86d/buffer.pyx#L231-L417 is deliberately designed to have a low-level but fast API, avoiding to creating Message or buffer slice objects.

protocol/codecs/base_scalar.go
fmoor: If we did that the buffer could not be reused because the memory would be shared with the calling code.

protocol/buff/reader.go
fmoor: 2. No. New slice objects are only created for new variables. Slices are the equivalent to `struct { Data uintptr, Len int, Cap int }` and slice operations only update the field values same as a buffer implementation would do.

1. Yes. All but one alocation could be eliminated by reusing the Message object.

fmoor:
> I worry that the read buffer is a bit more high level than it should be.

I have micro benchmarks in the test suite that I have been watching closely. The big number is the iteration count.

```bash
$ go test -bench=Pop -benchmem github.com/edgedb/edgedb-go/protocol/buff
goos: linux
goarch: amd64
pkg: github.com/edgedb/edgedb-go/protocol/buff
BenchmarkPopUint8-8     1000000000               0.381 ns/op           0 B/op          0 allocs/op
BenchmarkPopUint16-8    1000000000               0.488 ns/op           0 B/op          0 allocs/op
BenchmarkPopUint32-8    1000000000               0.457 ns/op           0 B/op          0 allocs/op
BenchmarkPopUint64-8    198831166                5.87 ns/op            0 B/op          0 allocs/op
BenchmarkPopUUID-8      362758321                3.32 ns/op            0 B/op          0 allocs/op
BenchmarkPopBytes-8     127474308                9.50 ns/op            0 B/op          0 allocs/op
BenchmarkPopString-8    77797806                15.1 ns/op             0 B/op          0 allocs/op
```

protocol/codecs/base_scalar.go
fmoor: I think I misunderstood the question. I intentionally kept this change to the buffer. I plan to rip out reflect.Value from the Decoder.Decode() call and replace it with unsafe.Pointer every where. This should be a significant speed up.

PR #18, opened 2020-11-17
decode into unsafe.Pointer instead of reflect.Value

fmoor:
# After
image

# Before
image

### DIFF
--- a/protocol/buff/reader.go
+++ b/protocol/buff/reader.go
@@ -25,19 +25,19 @@ import (
 
 // Message is a Reader with a Type attribute.
 type Message struct {
-	bts  []byte
+	Bts  []byte
 	Type uint8
 }
 
 // NewMessage returns a new Message.
 func NewMessage(bts []byte) *Message {
-	return &Message{bts: bts}
+	return &Message{Bts: bts}
 }
 
 // Finish asserts that the message has been fully read.
 // It panics if it has not.
 func (b *Message) Finish() {
-	if len(b.bts) > 0 {
+	if len(b.Bts) > 0 {
 		panic(fmt.Sprintf(
 			"cannot finish: unread data in buffer (message type: 0x%x)",
 			b.Type,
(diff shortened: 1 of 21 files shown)


protocol/buff/reader.go
```diff
@@ -25,19 +25,19 @@ import (
 
 // Message is a Reader with a Type attribute.
 type Message struct {
-	bts  []byte
+	Bts  []byte
```
1st1, line 30: I'd call it `Payload`, but I hope we'll be able to remove this struct in future PRs.

protocol/codecs/base_scalar.go
```diff
@@ -209,10 +209,15 @@ func (c *Bytes) Type() reflect.Type {
...
-	// todo use out's memory if it has enough allocated :thinking:
-	o := make([]byte, len(b))
-	copy(o, b)
-	out.SetBytes(o)
+	p := (*[]byte)(unsafe.Pointer(out.UnsafeAddr()))
+
+	if cap(*p) >= len(b) {
+		*p = (*p)[:len(b)]
+	} else {
+		*p = make([]byte, len(b))
```
1st1: What are the scenarios when `out` might not have enough memory pre-allocated?

protocol/codecs/base_scalar.go
```diff
@@ -138,8 +138,13 @@ func (c *UUID) Type() reflect.Type {
...
 // Decode a UUID.
 func (c *UUID) Decode(msg *buff.Message, out reflect.Value) {
+	if len(msg.Bts) < 20 {
+		panic("buffer overread")
+	}
+
 	p := (*types.UUID)(unsafe.Pointer(out.UnsafeAddr()))
-	copy((*p)[:], msg.PopBytes())
+	copy((*p)[:], msg.Bts[4:20])
+	msg.Bts = msg.Bts[20:]
```
1st1, line 145: Oh, it would be great if you could squash related commits before the review next time.

protocol/codecs/base_scalar.go
fmoor: Out has zero bytes allocated.

```go
var out []byte
con.QueryOneJson(ctx, query, &out)
```

The same thing happens if a struct with a slice field is allocated this way. The slice gets zero bytes.

protocol/buff/reader.go
```diff
@@ -50,32 +50,30 @@ func (b *Message) Len() int {
 	return len(b.Bts)
 }
 
-// Discard skips the next n bytes.
-func (b *Message) Discard(n int) {
+// AssertAllocated panics if there aren't n bytes in the buffer.
+func (b *Message) AssertAllocated(n int) {
 	if len(b.Bts) < n {
 		panic("buffer overread")
```
1st1, line 64: Would be great if this function is inlined. But AFAIK Go doesn't inline functions that can panic. I wonder if it does, can you check that?

And if it doesn't, maybe we can trick the compiler to do it anyways if we wrap `panic()` into our own helper function.

1st1 (approved):
Awesome work.

protocol/buff/reader.go
fmoor: It does inline this function.

```bash
$ go build -gcflags=-m github.com/edgedb/edgedb-go/protocol/codecs 2>&1 | grep AssertAllocated
protocol/codecs/array.go:73:13: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/array.go:77:14: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/array.go:81:13: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/base_scalar.go:142:21: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/base_scalar.go:215:21: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/base_scalar.go:259:13: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/object.go:109:13: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/object.go:112:14: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/object.go:118:15: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/object.go:121:16: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/set.go:68:13: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/set.go:72:14: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/set.go:76:13: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/tuple.go:88:13: inlining call to buff.(*Message).AssertAllocated
protocol/codecs/tuple.go:94:14: inlining call to buff.(*Message).AssertAllocated
```

1st1: Cool, thanks for confirming

protocol/buff/reader.go
fmoor: I was thinking of merging Message into the writer so that there is just a buff.Buff

PR #20, opened 2020-11-17
merge read and write buffers into one buffer type

fmoor:


protocol/buff/buff.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// PeekUint32 reads a uint32 but does not advance the buffer.
+func (b *Buff) PeekUint32() uint32 {
+	return binary.BigEndian.Uint32(b.Msg)
+}
+
+// PopUint64 reads a uint64 and advances the buffer.
+func (b *Buff) PopUint64() uint64 {
+	val := binary.BigEndian.Uint64(b.Msg[:8])
```
1st1, line 262: Will it complain if `b.Msg` has less than 8 bytes in it?

protocol/buff/buff.go
```diff
@@ -0,0 +1,288 @@
...
+			panic("buffer overread")
+		}
+
+		pos := 1 + int(binary.BigEndian.Uint32(b.payload[b.mPos+1:b.mPos+5]))
+
+		if len(b.payload) < b.mPos+pos {
+			panic("buffer overread")
+		}
+
+		b.Msg = b.payload[b.mPos+5 : b.mPos+pos]
```
1st1, line 191: I assume we copy data here, is that right? Or this is like a "memory view"?

protocol/buff/buff.go
fmoor: yes it will panic

protocol/buff/buff.go
fmoor: slices in go are a view into an array they are [literally](https://golang.org/pkg/reflect/#SliceHeader)
```go
type SliceHeader struct {
    Data unsafe.Pointer
    Len  int
    Cap  int
}
```

Assigning to a slice only sets the three fields. It does not copy the underlying data.

protocol/buff/buff.go
1st1: Good.

protocol/buff/buff.go
1st1: Awesome.

1st1 (approved):
Great work, Frederick. I wonder if benchmarks see any improvements with this.

fmoor:
> I wonder if benchmarks see any improvements with this.

The difference is less than the noise in this benchmark. Profiling the benchmarks though shows that there is a measurable speedup by removing the PopMessage function.

# After 
image

# Before 
image

PR #21, opened 2020-11-18
persist buffer with connection

fmoor:
Keep a buffer per connection to reduce allocations.

### DIFF
--- a/granular_flow.go
+++ b/granular_flow.go
@@ -83,7 +83,7 @@ func (c *Client) pesimistic(
 	q query,
 	tp reflect.Type,
 ) error {
-	ids, err := prepare(ctx, conn, q)
+	ids, err := c.prepare(ctx, conn, q)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


granular_flow.go
```diff
@@ -222,47 +222,47 @@ func (c *Client) execute(
 	tp reflect.Type,
 	cdcs codecPair,
 ) error {
-	buf := buff.New(c.buffer[:0])
-	buf.BeginMessage(message.Execute)
-	buf.PushUint16(0)       // no headers
-	buf.PushBytes([]byte{}) // no statement name
-	cdcs.in.Encode(buf, q.args)
-	buf.EndMessage()
+	c.buf.Reset()
```
1st1, line 225: Hm.... is this safe? What if there are some unprocessed (asynchronously received) messages in the buf? We'll have to deal with those soon. 

I don't think using a new buffer each time you want to compose a message is a noticeable performance hit, but it's certainly safer and easier to reason about.

fmoor: This change is not meant to be the long term strategy. It is tidying up some inconsistencies. Previously prepare allocated on every call, while optimistic reused reallocated memory. 

For now this is safe because buffers are not used concurrently, even when they are used concurrently it is easy to ensure that only one go routine has access to a given buffer using channels.

1st1: If this makes refactoring is temporary and makes your life easier - then sure. But we'll need to revise this.

1st1: Although I still feel very uneasy about this. It would make more sense to add two separate buffers to Client: `readBuf` and `writeBuf`

fmoor:
scratching this change

Issue #19, opened 2020-11-17
Pooling implementation


fmoor:
per https://github.com/edgedb/edgedb-go/pull/1#discussion_r519968141

## Pool Configuration
Pool configuration is fixed after creation.
- `MaxConns`, default = `max(4, runtime.NumCPU())`
- `MinConns`, default = 1

## Functions
```go
func Connect(context.Context, Options) (*Pool, error)

// ignores pool specific configuration
func ConnectOne(context.Context, Options) (*Conn, error)
```

## Pool API
```go
type Pool struct {
    func (*Pool) Execute(context.Context, string) error
    func (*Pool) Query(context.Context, string, interface{}, ...interface{}) error
    func (*Pool) QueryOne(context.Context, string, interface{}, ...interface{}) error
    func (*Pool) QueryJSON(context.Context, string, *[]byte, ...interface{}) error
    func (*Pool) QueryOneJSON(context.Context, string, *[]byte, ...interface{}) error
    func (*Pool) Acquire(context.Context) (*PoolConn, err)

    // Close closes all connections blocking until acquired connections are released.
    func (*Pool) Close() error
}
```

## Conn API
```go
type baseConn struct {
    func (*Conn) Execute(context.Context, string) error
    func (*Conn) Query(context.Context, string, interface{}, ...interface{}) error
    func (*Conn) QueryOne(context.Context, string, interface{}, ...interface{}) error
    func (*Conn) QueryJSON(context.Context, string, *[]byte, ...interface{}) error
    func (*Conn) QueryOneJSON(context.Context, string, *[]byte, ...interface{}) error
}

type Conn struct {
    Conn // i.e. has all the methods Conn does
    func (*Conn) Close() error
}

type PoolConn struct {
    Conn // i.e. has all the methods Conn does

    // Release puts the connection back in the Pool.
    // After the first call to Release() all methods on PoolConn will panic when called.
    func (*PoolConn) Release()
}
```

fmoor:
@1st1 let me know if I missed anything important here.

1st1:
> MinConns, default = 0

What does this mean? Ideally, when you create the pool, it should establish at least one connection (to verify connection params etc).

> MaxConnIdleTime and MaxConnLifeTime

We'll actually have these tuneable at the DB level (on the server side). Having them on the client side is great, but maybe we'll need to make sure that it's not possible to set up `MaxConnIdleTime` to be longer on the client than it is configured on the server. cc @tailhook and @elprans 

[...]

PR #28, opened 2020-11-24
implement optional pooling

fmoor:
This change adds:
- the ability to connect either a single connection or a pool
- the ability to acquire connections from a pool
- termination of all connections when a pool is closed
- configurable max and min connection counts

fixes https://github.com/edgedb/edgedb-go/issues/19

## After
image

## Before
image

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -27,21 +27,14 @@ import (
 
 // Pool is a pool of connections.
 type Pool struct {
-	// mu locks the closeSignal channel in calls to Pool.Close()
-	mu sync.Mutex
-
-	// pool.Close() sends on this channel to signal closing the pool
-	closeSignal chan chan struct{}
+	isClosed bool
+	mu       sync.RWMutex // locks isClosed
 
 	// A buffered channel of connections ready for use.
 	freeConns chan *baseConn
 
 	// A buffered channel of structs representing unconnected capacity.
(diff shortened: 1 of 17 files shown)


pool.go
```diff
@@ -0,0 +1,349 @@
...
+
+	opts          *Options
+	typeIDCache   *cache.Cache
+	inCodecCache  *cache.Cache
+	outCodecCache *cache.Cache
+}
+
+// Connect a pool of connections to a server.
+func Connect(ctx context.Context, opts Options) (*Pool, error) { // nolint
+	// todo should 0 be a valid value for MinConns?
```
1st1, line 54: No, I don't think 0 makes sense.

1st1 (approved):

fmoor:
I remove background go-routine from pool implementation.

The background goroutine only piped one channel into another and served to lock the "isClosed" pool state. In almost all real world scenarios this design would result in the background goroutine waking up, doing almost nothing, then sleeping again every time a connection was released. The overhead off waking a goroutine is nontrivial, so removing this background goroutine and using a read/write lock to guard the isClosed state improved the acquire/release performance by ~10x.

pool.go
fmoor: removed todo :heavy_check_mark:

1st1 (approved):

Issue #22, opened 2020-11-24
connect options feature parity with other edgedb drivers


fmoor:
```go
type Options struct {
    Hosts    []string
    Ports    []int
    User     string
    Database string
    Password string

    // Timeout is used for reconnecting in the background. 
    // The provided ctx argument's deadline is used for connecting in the foreground.
    Timeout time.Duration
}

func Connect(ctx context.Context, opts Options) (*Pool, error)
func ConnectDSN(ctx context.Context, dsn string, opts Options) (*Pool, error)

func ConnectOne(ctx context.Context, opts Options) (*Conn, error)
func ConnectOneDSN(ctx context.Context, dsn string, opts Options) (*Conn, error)
```

Issue #24, opened 2020-11-24
Error API


fmoor:
Careful consideration should be given to all returned errors and panic values. 

## References
https://go.googlesource.com/proposal/+/master/design/go2draft.md
https://pkg.go.dev/github.com/spacemonkeygo/errors
https://github.com/juju/errgo
https://pkg.go.dev/github.com/hashicorp/errwrap
https://pkg.go.dev/github.com/pkg/errors

1st1:
Yes, at least for the next couple of months until we implement https://github.com/edgedb/edgedb/discussions/1849 ;)

fmoor:
:question: What error type should checks that ought never fail return? (noting this here so I don't forget when I get around to implementing)

PR #31, opened 2020-12-01
support similar connect options as other clients

fmoor:
This change adopts a connect API similar to the java script client accepting a dsn string and an options object.

fixes https://github.com/edgedb/edgedb-go/issues/22


fmoor:
@1st1 If environment variables change after the initial call to connect and then a pool tries to make a new connection, should the environment variables be reread? or should it use the original values of environment variables for all time?

options.go
```diff
@@ -17,77 +17,420 @@
...
+
+	if val := os.Getenv("EDGEDB_DATABASE"); database == "" {
+		database = val
+	}
+
+	if database == "" {
+		database = user
+	}
+
+	if user == "" {
```
1st1, line 367: The default user/database should be "edgedb" -- see https://github.com/edgedb/edgedb-python/pull/131

options_test.go
```diff
@@ -18,67 +18,397 @@ package edgedb
...
+		if err != nil {
+			panic(err)
+		}
+	}
 }
 
-func TestDialHost(t *testing.T) {
-	opts := Options{Host: "some.com", Port: 1234}
-	assert.Equal(t, "some.com:1234", opts.address())
+func TestConUtils(t *testing.T) {
```
1st1, line 67: Did you copy these from edgedb-python?

options.go
```diff
@@ -17,77 +17,420 @@
...
+	var addrs []netParameters
+	for i := 0; i < len(hosts); i++ {
+		h := hosts[i]
+		p := ports[i]
+
+		if strings.HasPrefix(h, "/") {
+			if !strings.Contains(h, ".s.EDGEDB.") {
+				var sockName string
+				if opts.admin {
+					sockName = fmt.Sprintf(".s.EDGEDB.admin.%v", p)
```
1st1, line 387: I'm not sure if we should support admin sockets in Go driver. @elprans ?

1st1 (approved):
Overall looks good. I'm wondering, did you fully replicate the functionality/logic/tests from edgedb-python & edgedb-js?

options.go
elprans: Admin sockets are deprecated across the board, please don't add support.

fmoor: The admin option is not part of the public API. It is only there right now for running tests in CI.

1st1: > The admin option is not part of the public API. It is only there right now for running tests in CI.

I suggest you to take a look how edgedb-js CI is set up. There's no need for admin sockets that way (at least it shouldn't be).

I'd prefer not adding them in any API, public or private.

fmoor: I tried removing the admin option and then figuring out how to make the tests connect to a server created with the `edgedb-server` command. I can't figure out what the default password is for either the `edgedb` user or the linux user. The edgedb-python test suite seems to be using the admin socket to set a password for the `edgedb` user.

https://github.com/edgedb/edgedb-python/blob/289592fb540d2f4257956948823a309596bbe058/edgedb/_testbase.py#L70
https://github.com/edgedb/edgedb-python/blob/289592fb540d2f4257956948823a309596bbe058/edgedb/_cluster.py#L242
https://github.com/edgedb/edgedb-python/blob/289592fb540d2f4257956948823a309596bbe058/edgedb/_cluster.py#L228

Is there something I am missing?

fmoor: I am running edgedb-server like this
```bash
edgedb-server-1-alpha6 -D /tmp/edgedb-go-test-server-006711182 --testmode --runstate-dir /tmp/edgedb-go-test-server-006711182 --port 49697
```

1st1: hmm... @elprans ?

elprans: You also want 

```
--bootstrap-script='CONFIGURE SYSTEM INSERT Auth {priority := 0, method := (INSERT Trust)}`
```

elprans: **Or** (better):

```
--bootstrap-script='CREATE SUPERUSER ROLE test { SET password := "test-password" }'
```

fmoor: @elprans it doesn't look like `--bootstrap-script` is an option :confused: 

```bash
$ edgedb-server-1-alpha6 --temp-dir --testmode --echo-runtime-info --port=auto --auto-shutdown --bootstrap-script='CREATE SUPERUSER ROLE test { SET password := "shhh"  }'
Usage: edgedb-server-1-alpha6 [OPTIONS]
Try 'edgedb-server-1-alpha6 -h' for help.

Error: no such option: --bootstrap-script
```

```bash
$ edgedb-server-1-alpha6 -h
Usage: edgedb-server-1-alpha6 [OPTIONS]

Options:
  -D, --data-dir PATH             database cluster directory
  --postgres-dsn TEXT             DSN of a remote Postgres cluster, if using
                                  one

  -l, --log-level TEXT            Logging level.  Possible values: (d)ebug,
                                  (i)nfo, (w)arn, (e)rror, (s)ilent

  --log-to DEST                   send logs to DEST, where DEST can be a file
                                  name, "syslog", or "stderr"

  --bootstrap                     bootstrap the database cluster and exit
  --default-database TEXT         the name of the default database to create
  --default-database-user TEXT    the name of the default database owner
  --devmode / --no-devmode        enable or disable the development mode
  --testmode / --no-testmode      enable or disable the test mode
  -I, --bind-address TEXT         IP address to listen on
  -P, --port PORT                 port to listen on
  -b, --background                daemonize
  --pidfile-dir PATH              path to PID file directory
  --daemon-user INTEGER
  --daemon-group INTEGER
  --runstate-dir PATH             directory where UNIX sockets and other
                                  temporary runtime files will be placed
                                  (/run/edgedb by default)

  --max-backend-connections INTEGER
  --echo-runtime-info             echo runtime info to stdout; the format is
                                  JSON, prefixed by "EDGEDB_SERVER_DATA:",
                                  ended with a new line

  --temp-dir                      create a temporary database cluster
                                  directory that will be automatically purged
                                  on server shutdown

  --auto-shutdown                 shutdown the server after the last
                                  management connection is closed

  --max-protocol TEXT             maximum supported and advertized client
                                  protocol version

  --version                       Show the version and exit.
  -h, --help                      Show this message and exit.
```

elprans: Oh, that's in nightly only, and it's actually `--bootstrap-command` (there's also `--bootstrap-script` taking a file path).

fmoor: I confirmed that I can make this work on a nightly build. Is everyone comfortable with the CI for this repo running against a nightly build? tentatively `edgedb-server-1-alpha6-dev5104`

elprans: >  Is everyone comfortable with the CI for this repo running against a nightly build

Yes

> `edgedb-server-1-alpha6-dev5104`

This seems old.  `edgedb server install --nightly` reports `1.0a7.dev5243+g096c48c9-2020120300~focal.nightly`

fmoor: ok, using `edgedb-server-1-alpha7-dev5249`

options_test.go
fmoor: This was an attempt to copy what was in edgedb-python. I went back and found that I had several errors. I also compared this with edgedb-js to be sure there wasn't anything different there.

fmoor:
> did you fully replicate the functionality/logic/tests from edgedb-python & edgedb-js?

I believe that I have fully replicated the logic for parsing connect arguments and trying each address in turn until one of them connects successfully or all addresses fail. There are a few differences from edgedb-python/js:

- deprecated features were omitted in this implementation.
- js always uses credentials file user & port, py uses credentials file user & port only if no user or port arguments were passed to `connect()`. This implementation follows edgedb-python.
- the exposed connect API is:
```go
func Connect(ctx context.Context, opts Options) (*Pool, error)
func ConnectDSN(ctx context.Context, dsn string, opts Options) (*Pool, error)

func ConnectOne(ctx context.Context, opts Options) (*Conn, error)
func ConnectOneDSN(ctx context.Context, dsn string, opts Options) (*Conn, error)
```

I'm not sure I love this connect API. I'm not sure the `DSN` variants are necessary. Also, I almost want to change the whole thing to:
```
func DSN(dsn string) (Options, err)
func Instance(name string) (Options, err) // or maybe call this one CredentialsFile :thinking: 

func Connect(ctx context.Context, opts Options) (*Pool, error)
func ConnectOne(ctx context.Context, opts Options) (*Conn, error)
```

fmoor:
[pgx](https://pkg.go.dev/github.com/jackc/pgx#ParseConnectionString) has several functions that all return the equivalent to `Options` then connect just takes an `Options`. This pattern seems common in go. However, this change would leave a lot of work up to the user if they wanted to merge a dsn or credential file with their own defaults the way that the current api allows or a builder API would need to be added to `Options`. I don't have a strong sense of which would be better :thinking:

options_test.go
1st1: Yeah, that's awesome. I was only asking because I *thought* the implementation there was good tested :)

1st1 (approved):

options.go
```diff
@@ -17,80 +17,74 @@
...
+	//
+	// Ports may either be:
+	//
+	// - the same length ans Hosts
+	//
+	// - a single port to be used all specified hosts
+	//
+	// - empty indicating the value parsed from the dsn argument
+	//   should be used, or the value of the EDGEDB_PORT environment variable,
+	//   or 5656 if neither is specified.
```
tailhook, line 56: Not sure I understand structure here. Is this "Hosts <crossproduct> Ports" or parallel arrays where first host in Hosts corresponds to first port in ports?

What are the downsides of this being an array of pairs?

credentials.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return nil, fmt.Errorf(
+			"cannot read credentials at %q: %v%w", path, err, ErrBadConfig,
+		)
+	}
+
+	return creds, nil
+}
+
+func validateCredentials(data map[string]interface{}) (*credentials, error) {
```
tailhook, line 59: I see an issue with defaults here. E.g. default user and database are `edgedb`, default port is 5656.

The problem with applying these defaults later is that, config like this:
```
{ database: "" }
```
Would be treated differently in no an non go. In all other languages it's the error (database name is empty). But because golang doesn't distinguish between empty string and non-existing string, it will be equivalent to `database: "edgedb"`.

options.go
```diff
@@ -17,80 +17,74 @@
...
 import (
-	"fmt"
-	"net/url"
-	"strconv"
-	"strings"
+	"time"
 )
 
 // Options for connecting to an EdgeDB server
 type Options struct {
```
tailhook, line 24: Eventually, we want to hide these as private fields and use builder/fluent interface. I.e. only expose methods I think. Not sure if this has to be the same PR.

tailhook (commented):
Two big things:
1. I don't see why we need `Options` and `connConfig`. I would expose latter, with nice methods to create and modify the structure. Like this in Rust: https://github.com/edgedb/edgedb-rust/blob/master/edgedb-client/src/builder.rs#L90-L206 Because `Options` is a bit confusing and not forward-compatible

2. I also have doubts with errors. Currently you use a bunch of singletons as errors. This is a known pattern in Go, but the problem is that not all of the errors could be singletons. So users have to learn which are constants and use `errors.Is` with them and which are types and use `errors.As` with them. And also can't switch from singleton to type if some error has two different messages now. Is this what is considered normal for gophers?

fmoor:
>     1. I don't see why we need `Options` and `connConfig`. I would expose latter, with nice methods to create and modify the structure. Like this in Rust: https://github.com/edgedb/edgedb-rust/blob/master/edgedb-client/src/builder.rs#L90-L206 Because `Options` is a bit confusing and not forward-compatible

It was not clear to me that the builder approach was the accepted and final future design, so I opted to follow the current py & js implementations. If the builder approach is finalized I'm happy to change this PR. If it is not finalized yet, I'd prefer to merge this and rework it when it is finalized.

>     2. I also have doubts with errors. Currently you use a bunch of singletons as errors. This is a known pattern in Go, but the problem is that not all of the errors could be singletons. So users have to learn which are constants and use `errors.Is` with them and which are types and use `errors.As` with them. And also can't switch from singleton to type if some error has two different messages now. Is this what is considered normal for gophers?

This PR is not intended to provide a coherent error API. There is  an issue for that here https://github.com/edgedb/edgedb-go/issues/24

options.go
fmoor: This was an attempt to follow the js driver behavior. The builder pattern would be much a nicer API in go it just didn't seem like that design was finalized. I'm just trying to get off the ground with this PR.

1st1:
> This PR is not intended to provide a coherent error API. There is an issue for that here #24

Yeah, I'd merge this as is unless there are other objections and continue iterating on the API in follow up PRs.

Issue #23, opened 2020-11-24
handle sockets asyncronusly 


fmoor:
All connected sockets must respond immediately to server initiated messages. 

Consider testing socket reading with two types of functional tests.
1. test huge messages with a known message query maybe string repeat
2. random message sizes

PR #37, opened 2020-12-16
listen on socket when not querying

fmoor:
Adds processing server initiated messages in a background go routine when queries are not in progress.

fixes https://github.com/edgedb/edgedb-go/issues/23

Benchmark results:
image


edgedb.go
```diff
@@ -54,29 +62,98 @@ func ConnectOne(ctx context.Context, opts Options) (*Conn, error) { // nolint:go
...
+	deadline, _ := ctx.Deadline()
+	if e := c.conn.SetDeadline(deadline); e != nil {
+		return fmt.Errorf("%v", e)
+	}
+
+	return nil
+}
+
+func (c *baseConn) acquireReader(ctx context.Context) (*buff.Reader, error) {
+	if e := c.conn.SetReadDeadline(time.Now()); e != nil {
```
fmoor, line 101: I initially implemented a background go routine that read from the socket in a tight loop. That started getting messy when it came to recovering from errors. This implementation starts a background reader when the `client.QueryX()` call ends and stops it when a new query begins. The down side to this is that the only reliable way to get the reader out of the background is to make it timeout. Do you think this is an acceptable compromise on complexity @1st1?

1st1: > The down side to this is that the only reliable way to get the reader out of the background is to make it timeout.

Hm, this sounds really suboptimal. Is it possible to push some sentinel value to wake up the listener goroutine blocked on `buff.Reader.Next()`?

1st1: Actually I'm not sure if it's possible to implement what I suggested.

> I initially implemented a background go routine that read from the socket in a tight loop. That started getting messy when it came to recovering from errors. 

Can you elaborate on the messy part? Because having a tight loop goroutine to read/parse messages and push signals to channels would make everything else easier.

Basically, you'd have a ReadBuffer where you'd feed data in that tight loop reader goroutine. Whenever there's a new complete message in that buffer you'd push `true` to the channel. 

This way, in say `Query()` you will just try reading directly from `ReadBuffer` to see if there's data in it. If not, you'd wait on the channel until the data is available.

And for asynchronous parsing you'd have yet another goroutine just listening on two channels: one for the ReadBuffer, and another would be used to stop it from the next `Query()` call.

The first goroutine (the one that populates ReadBuffer) can be stopped by closing the socket.

fmoor: I'll roll back to my original implementation and see if I can get that working.

This is not my original concern but loosely related: What happens when a query times out part way through reading data from the socket? The next read from the socket is going to return the remaining part of the timed out message(s). How does the client recover in this scenario? Close the connection and reconnect?

1st1: > What happens when a query times out part way through reading data from the socket?

We will soon have an ability to enforce timeouts on the server side for every query. That way, if the query times out on the client side, the client ideally should wait a bit longer to try to read all messages until `SYNC`.  Or else it should just terminate the connection and open a new one.

1st1: > I'll roll back to my original implementation and see if I can get that working.

Thanks! I think it should be doable; maybe some parts will be tricky but the overall solution in the end should be fast and intuitive.

fmoor: updated with background reader go routine

edgedb.go
```diff
@@ -74,36 +81,98 @@ func connectOne(ctx context.Context, cfg *connConfig, conn *baseConn) error {
...
+		conn.readerChan = make(chan *buff.Reader, 1)
+		conn.releaseReader(r)
 		return nil
 	}
 
 	conn.conn = nil
 	return err
 }
 
+func (c *baseConn) setDeadline(ctx context.Context) error {
```
1st1, line 119: Hm, what's this function for?

edgedb.go
```diff
@@ -74,36 +81,98 @@ func connectOne(ctx context.Context, cfg *connConfig, conn *baseConn) error {
...
+		return nil, ctx.Err()
+	}
+}
+
+func (c *baseConn) releaseReader(r *buff.Reader) {
+	go func() {
+		for r.Next(c.acquireReaderSignal) {
+			if e := c.fallThrough(r); e != nil {
+				// todo what is the right thing to do here?
+				panic(e)
```
1st1, line 148: The right thing would be to set some internal flag "client.failed" to "true" and close all IO. Then error out to the user when they try to use any client API that the client has been closed because of an unexpected message in the protocol.

edgedb.go
1st1: Ah, that's for client-side timeouts, right?

script_flow.go
```diff
@@ -17,46 +17,43 @@
...
-		switch buf.MsgType {
+	var err error
+	done := buff.NewSignal()
+
+	for r.Next(done.Chan) {
+		switch r.MsgType {
 		case message.CommandComplete:
-			buf.PopUint16() // header count (assume 0)
-			buf.PopBytes()  // command status
+			r.Discard(2) // header count (assume 0)
```
1st1, line 40: Please add an internal function similar to `reject_headers` in edgedb-python

internal/soc/soc.go
```diff
@@ -0,0 +1,80 @@
...
+			freeMemory.Release(slab)
+		}
+	}
+
+	for {
+		slab := freeMemory.Acquire()
+		buf := slab
+
+		for len(buf) >= minChunkSize {
+			n, err := conn.Read(buf)
```
1st1, line 57: Why don't you check `err` right here?

It looks like this goroutine is always reading data, right? So at some time the socket connection can be closed, in which case we just need to exit immediately, basically with check like this:


```
if err != nil {
  if conn.closed {
    return
  } else {
    toBeDeserialized <- &Data{Err: err}
    return
  }
}
```

edgedb.go
fmoor: correct :heavy_check_mark:

edgedb.go
fmoor: :+1: created new issue here https://github.com/edgedb/edgedb-go/issues/39

script_flow.go
fmoor: :+1: created issue here https://github.com/edgedb/edgedb-go/issues/40

internal/soc/soc.go
fmoor: if there is an error and the connection is closed, the upstream go routine needs to get the error too or it will block forever waiting for data.

This sends data before checking the error because [the go docs say](https://golang.org/pkg/io/#Reader):
> Callers should always process the n > 0 bytes returned before considering the error err.

connect.go
```diff
@@ -17,80 +17,96 @@
...
 			if major != 0 || minor != 8 {
-				if err := c.conn.Close(); err != nil {
-					return err
-				}
-
-				return fmt.Errorf(
-					"unsupported protocol version: %v.%v",
-					major,
-					minor,
+				e := fmt.Errorf(
```
1st1, line 62: While here I'd make the min (or max?) supported protocol version a constant.

connect.go
```diff
@@ -17,80 +17,96 @@
...
+					"unsupported protocol version: %v.%v", major, minor,
 				)
+				return wrapAll(err, e, c.conn.Close())
 			}
 		case message.ServerKeyData:
-			buf.Discard(32) // key data
+			r.Discard(32) // key data
```
1st1, line 68: I recommend implementing `discard_message`-like API. https://github.com/MagicStack/py-pgproto/blob/90ad4154a0aa6923e2f97d3d3d779c4b830ff415/buffer.pyx#L639-L655

BTW, what happens if I don't consume message data to its end? Or overread it?

edgedb.go
```diff
@@ -74,36 +81,98 @@ func connectOne(ctx context.Context, cfg *connConfig, conn *baseConn) error {
...
-		err = conn.connect(ctx, cfg)
+		err = conn.connect(r, cfg)
 		if err != nil {
-			_ = conn.conn.Close()
-			log.Printf("while attempting connection %+v: %+v", addr, err)
+			err = wrapAll(err, conn.conn.Close())
 			continue
 		}
 
+		conn.acquireReaderSignal = make(chan struct{}, 1)
```
1st1, line 109: I'm wondering how creating a channel for `struct{}` is better than creating one for, say, booleans here? IMO sending only `struct{}{}` over it looks weird.

edgedb.go
```diff
@@ -74,36 +81,98 @@ func connectOne(ctx context.Context, cfg *connConfig, conn *baseConn) error {
...
 }
 
 // Execute an EdgeQL command (or commands).
-func (c *baseConn) Execute(ctx context.Context, cmd string) (err error) {
-	return c.scriptFlow(ctx, c.conn, cmd)
+func (c *baseConn) Execute(ctx context.Context, cmd string) error {
+	r, err := c.acquireReader(ctx)
+	if err != nil {
+		return err
+	}
```
1st1, line 167: shouldn't you `defer c.releaseReader(r)` right here?

granular_flow.go
```diff
@@ -41,227 +39,257 @@ func (c *baseConn) granularFlow(
...
+
+	for r.Next(done.Chan) {
+		switch r.MsgType {
 		case message.CommandDataDescription:
-			buf.PopUint16() // num headers is always 0
-			buf.PopUint8()  // cardianlity
+			// num headers is always 0
+			// cardianlity
+			// input descriptor ID
+			r.Discard(19)
```
1st1, line 194: Why are you discarding descriptor IDs?

granular_flow.go
```diff
@@ -41,227 +39,257 @@ func (c *baseConn) granularFlow(
...
-	for buf.Next() {
-		switch buf.MsgType {
+	err := ErrZeroResults
+	done := buff.NewSignal()
+
+	for r.Next(done.Chan) {
+		switch r.MsgType {
 		case message.Data:
-			buf.PopUint16() // number of data elements (always 1)
+			r.Discard(2) // number of data elements (always 1)
```
1st1, line 253: Let's actually read data and put real assertions in places like this.

granular_flow.go
```diff
@@ -41,227 +39,257 @@ func (c *baseConn) granularFlow(
...
-			buf.PopUint16() // header count (assume 0)
-			buf.PopBytes()  // command status
+			r.Discard(2) // header count (assume 0)
+			r.PopBytes() // command status
 		case message.ReadyForCommand:
-			buf.PopUint16() // header count (assume 0)
-			buf.PopUint8()  // transaction state
+			// header count (assume 0)
+			// transaction state
+			r.Discard(3)
```
1st1, line 272: Need a helper for reading transaction state. The latest one should be stored in a private field on the connection object. We'll need to add public APIs to allow users to see it (not sure if we support that in JS/Python, but we should).

internal/buff/read.go
```diff
@@ -0,0 +1,226 @@
...
+			if r.data.Err != nil {
+				r.Err = r.data.Err
+				r.data.Release()
+				r.data = nil
+				return false
+			}
+		}
+	}
+
+	r.Err = r.feed(5)
```
1st1, line 94: What does `5` mean here?

internal/buff/read.go
```diff
@@ -0,0 +1,226 @@
...
+// PopUint8 returns the next byte and advances the buffer.
+func (r *Reader) PopUint8() uint8 {
+	val := r.Buf[0]
+	r.Buf = r.Buf[1:]
+	return val
+}
+
+// PopUint16 reads a uint16 and advances the buffer.
+func (r *Reader) PopUint16() uint16 {
+	val := binary.BigEndian.Uint16(r.Buf[:2])
```
1st1, line 179: Will `r.Buf[:2]` panic if there's only one byte in `r.Buf`?

internal/buff/write.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	var n int
+	n, err = conn.Write(w.buf)
+	if err != nil {
+		return err
+	}
+
+	if n != len(w.buf) {
+		panic("not all data was sent")
```
1st1, line 59: Hm, why panic and not try sending the remaining data? Is writing always supposed to succeed and panicing is the recommended way to deal with it?

internal/buff/write.go
```diff
@@ -0,0 +1,167 @@
...
+import (
+	"encoding/binary"
+	"io"
+
+	"github.com/edgedb/edgedb-go/types"
+)
+
+// Writer is a write buffer.
+type Writer struct {
+	alocatedMemory [256 * 1024]byte
```
1st1, line 28: This is a very big buffer. Our WriteBuffer in Python starts with just 1kb, which is actually plenty for most cases. It can grow on demand in chunk increments. We need to implement the same strategy here.

internal/buff/write.go
```diff
@@ -0,0 +1,167 @@
...
+// and allocates space for message length.
+// BeginMessage panics if the EndMessage was not called
+// for the previous message.
+func (w *Writer) BeginMessage(mType uint8) {
+	if w.msgPos != 0 {
+		panic("cannot begin message: the previous message is not finished")
+	}
+
+	w.msgPos = 1 + len(w.buf)
+	w.buf = append(w.buf, mType, 0, 0, 0, 0)
```
1st1, line 149: Is `append` a stdlib function? What will it do when you are at the end of `alocatedMemory`?

internal/codecs/base_scalar.go
```diff
@@ -524,21 +522,20 @@ func (c *Duration) Type() reflect.Type {
 }
 
 // Decode a duration.
-func (c *Duration) Decode(buf *buff.Buff, out unsafe.Pointer) {
-	buf.PopUint32() // data length
-	microseconds := int64(buf.PopUint64())
-	buf.PopUint32() // reserved
-	buf.PopUint32() // reserved
+func (c *Duration) Decode(r *buff.Reader, out unsafe.Pointer) {
+	r.Discard(4) // data length
```
1st1, line 526: Why are you discarding data length in all decoders? You should actually read it and enforce it in every codec. In Python we have the `FRBuffer` helper to do that very efficiently (and `ReadBuffer` in JS), worth taking a look.

edgedb.go
fmoor: No. The reader should not be released if the connection failed because there is nothing pushing data from the socket to the reader. What should be here but isn't (I over looked this) is a check to see if the socket read produced any errors. if there was an error the socket should be closed or if recoverable the `soc.Read` go routine should be restarted. 

If something panics in `scriptFlow` the socket should be closed and the `baseConn` object should be unusable. Panics should never happen, they indicate a flaw in the driver. (currently this may not be true because https://github.com/edgedb/edgedb-go/issues/24 is not closed yet).

connect.go
fmoor: > BTW, what happens if I don't consume message data to its end? Or overread it?

The reader will panic. The r.Buf []byte is alwaysed caped so that over reading panics. If `len(r.Buf) > 0` then r.Next() panics.

edgedb.go
fmoor: Empty structs are semantic in go.
https://dave.cheney.net/2014/03/25/the-empty-struct
https://medium.com/@l.peppoloni/how-to-improve-your-go-code-with-empty-structs-3bd0c66bc531

In this case I see two advantages to using `struct{}{}` here.
1. `struct{}{}` has no value so it is clear that what comes out of the channel doesn't mean anything except that some thing came out of the channel. If it was `chan bool` one might mistakenly assign meaning to a `true` value that is somehow different than a `false` value.
2. `struct{}{}` has a width of `0` so it doesn't take any memory where `bool` has size of `1`. This difference is trivial in this case.

granular_flow.go
fmoor: Because I already have them in scope. Perhaps there is something I'm over looking here though.

granular_flow.go
fmoor: :+1:  created issue https://github.com/edgedb/edgedb-go/issues/42

granular_flow.go
fmoor: :+1: created issue https://github.com/edgedb/edgedb-go/issues/43

internal/buff/read.go
fmoor: Put 5 bytes into `r.Buf` so that the message type and length can be read.

internal/buff/read.go
fmoor: yes

internal/buff/write.go
fmoor: No. Good catch. This is incomplete :confused:

internal/buff/write.go
fmoor: Ok. I was making up numbers. I'll change that. It already grows if needed :+1:

internal/buff/write.go
fmoor: Yes its stdlib. it will allocate a new underly array that is ~ `2 * len(w.buf)`, copy the data in `w.buf` to the new array followed by the 5 bytes listed and return a new slice into the new array that ends at the last new 0. I think this is similar to the way `list.append()` works in python :thinking:

internal/codecs/base_scalar.go
fmoor: They are discarded because it didn't matter on day 1. It makes sense to check theme though in real world scenarios. Created an issue for this https://github.com/edgedb/edgedb-go/issues/44

connect.go
fmoor: > I recommend implementing discard_message-like API

What benefit does that abstraction provide in this implementation? None of the checks/actions that the python implementation performs are required here.

Maybe this would be more maintainable if the length of the message changed in the future?
```go
r.Buf = r.Buf[len(r.Buf):]
```

fmoor: on second thought this probably be better
```go
r.Buf = nil
```

1st1: > What benefit does that abstraction provide in this implementation?

You wouldn't need to count bytes by hand.

> r.Buf = nil

This is OK, but `r.DiscardMessage()` is way clearer to the reader of your code. It immediately tells the reader that you want to ignore the rest of the message, while `r.Buf = nil` requires the reader to understand your code way deeper.

The method would also allow you to tweak the actual implementation later, for which `r.Buf = nil` might be suboptimal.

In other words, compilers are smart enough to inline short helper functions like the proposed `DiscardMessage()`; we should use that fact when we can to write readable code.

edgedb.go
1st1: Thanks for explaining! 🎓

edgedb.go
1st1: I don't understand. You implement literally what I suggested in all other methods like `QueryOne`, what makes `Execute` different?

granular_flow.go
1st1: > Because I already have them in scope. 

I don't follow what you meant here, please elaborate.

edgedb.go
fmoor: They are also wrong. I over looked the error handling in this PR. I'll update it and have you review again.

granular_flow.go
fmoor: Sorry it's not in scope in this function. The PrepareComplete response returns the descriptor IDs too. Those values are used higher up in the call stack with the descriptor values obtained here. Is this the wrong thing to do?

connect.go
fmoor: done :heavy_check_mark:

internal/buff/write.go
fmoor: done :heavy_check_mark:

internal/buff/write.go
fmoor: updated :heavy_check_mark:

connect.go
fmoor: done :heavy_check_mark:

edgedb.go
fmoor: Updated. I'm not 100% I have it right but I think its better than it was.

edgedb.go
```diff
@@ -127,12 +207,29 @@ func (c *baseConn) QueryOne(
...
 		return err
 	}
 
-	return nil
+	if e := c.setDeadline(ctx); e != nil {
+		return e
+	}
+
+	err = c.granularFlow(r, val, q)
+	if soc.IsPermenentNetErr(err) {
```
1st1, line 220:
```suggestion
	if soc.IsPermanentNetErr(err) {
```

edgedb.go
```diff
@@ -154,11 +251,28 @@ func (c *baseConn) Query(
...
+		_ = c.conn.Close()
+		c.conn = nil
+	} else {
+		if e := c.setDeadline(context.Background()); e != nil {
+			_ = c.conn.Close()
+			c.conn = nil
+			return wrapAll(err, e)
+		}
+		c.releaseReader(r)
+	}
```
1st1, line 274: Can you create a helper function for this snippet that's repeated in multiple places?

internal/buff/read.go
1st1: Can you add a comment? Also how do you know that the buffer actually has 5 bytes in it?

1st1 (approved):

edgedb.go
fmoor: I moved all of that into `releaseReader`

internal/buff/read.go
fmoor: At this point we know that the server has sent more data so we know that there is another message to process. If there is a message to process then there must be at least 5 bytes that the server has sent or will send. `feed` waits on data from the `soc.Read` go routine if the reader has fewer than 5 bytes.

comment added :heavy_check_mark:

Issue #33, opened 2020-12-04
argument types that don't match query should return error instead of panic


fmoor:
Currently this panics on the type assertion for  `1` & `1111`

```go
var res []interface{}
ctx := context.Background()
err := conn.QueryOne(ctx, "SELECT (<int16>$0 + <int16>$1,)", &res, 1, 1111)
```

fmoor:
This seems to have been fixed in https://github.com/edgedb/edgedb-go/pull/51

fmoor:
I spoke too soon...

1st1:
> I spoke too soon...

You saw it happening again?

fmoor:
> You saw it happening again?

sorry, I'm conflating two issues. The bug outlined in the description is fixed. There is another one fixed in https://github.com/edgedb/edgedb-go/pull/53 that I erroneously reopened this for.

PR #53, opened 2020-12-29
fix stale write buffer bug

fmoor:
I added a test to start fixing https://github.com/edgedb/edgedb-go/issues/33 and discovered that it was already fixed, but the new test unearthed a bug caused by reusing the write buffer. This change creates a new write buffer for every payload sent to the server.

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -31,18 +31,19 @@ const (
 )
 
 func (c *baseConn) connect(r *buff.Reader, cfg *connConfig) error {
-	c.writer.BeginMessage(message.ClientHandshake)
-	c.writer.PushUint16(0) // major version
-	c.writer.PushUint16(8) // minor version
-	c.writer.PushUint16(2) // number of parameters
-	c.writer.PushString("database")
-	c.writer.PushString(cfg.database)
-	c.writer.PushString("user")
-	c.writer.PushString(cfg.user)
-	c.writer.PushUint16(0) // no extensions
-	c.writer.EndMessage()
+	w := buff.NewWriter(c.writeMemory[:0])
+	w.BeginMessage(message.ClientHandshake)
+	w.PushUint16(0) // major version
+	w.PushUint16(8) // minor version
+	w.PushUint16(2) // number of parameters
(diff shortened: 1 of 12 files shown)


fmoor:
This PR is based on https://github.com/edgedb/edgedb-go/pull/51. WIP until https://github.com/edgedb/edgedb-go/pull/51 is merged.

connect.go
```diff
@@ -31,18 +31,19 @@ const (
...
-	c.writer.PushUint16(2) // number of parameters
-	c.writer.PushString("database")
-	c.writer.PushString(cfg.database)
-	c.writer.PushString("user")
-	c.writer.PushString(cfg.user)
-	c.writer.PushUint16(0) // no extensions
-	c.writer.EndMessage()
-
-	if err := c.writer.Send(c.conn); err != nil {
+	w := buff.NewWriter(c.writeMemory[:0])
```
1st1, line 34: What does `c.writeMemory[:0]` do? Returns a pointer to the first byte of the buffer?

1st1 (commented):
It seems that this PR is way more than just about adding one test, no?

connect.go
fmoor: Basically, yes. `NewWriter()` takes a `[]byte` with `len(bytes) == 0` but expects it to have allocated capacity. `c.writeMemory[:0]` creates a zero length slice with allocated capacity equal to the length of the `writeMemory` array.

fmoor:
> It seems that this PR is way more than just about adding one test, no?

Yea. I added a test as a starting point to fix https://github.com/edgedb/edgedb-go/issues/33 but it was already fixed, but adding the test exposed another bug caused reusing the write buffer instead of creating a new one for each set of messages sent to the server. Maybe it should be called fix stale write buffer bug?

1st1:
> Maybe it should be called fix stale write buffer bug?

Doesn't matter much how the PR is called, but ideally we should make individual commits for every bug/feature -- way easier this way to use git bisect and just follow the history.

Issue #54, opened 2020-12-29
tuples of non-scalars break decoding


fmoor:
The current decoding scheme requires that the entire out value memory layout is known when a codec is built from a descriptor. This works well for all edgedb types except for tuple*. In go the only sequence of arbitrary types is `[]interface{}`, but this type obfuscates the memory layout of any elements in the slice. One solution to this problem could be to maintain 2 methods of deserializing. The first would be the fast path that requires knowing the out value's memory layout. The second would use the reflect package to decode into unknown memory layouts.

Note that named tuples don't have this problem because they can be mapped to a `struct` instead of `[]interface{}`.

*I'm assuming here that the JSON type will be `[]byte`. If JSON should be decoded it has the same problem as tuples.

PR #58, opened 2021-01-08
support decoding non-scalar tuple elements

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/54

This change enables decoding tuple elements to default types.
- Object and named tuples decode to map[string]interface{}
- tuples to []interface{}
- scalars to their respective types
- array & set to slice of whatever the element type decodes to

### DIFF
--- a/internal/codecs/array.go
+++ b/internal/codecs/array.go
@@ -42,10 +42,15 @@ func popArrayCodec(
 
 // Array is an EdgeDB array type codec.
 type Array struct {
-	id         types.UUID
-	child      Codec
-	typ        reflect.Type
-	step       int
+	id    types.UUID
+	child Codec
+	typ   reflect.Type
+
+	// step is the element width in bytes for a go array of type `Array.typ`.
+	step int
+
+	// useReflect indicates weather reflection or a known memory layout
+	// should be used to deserialize data.
 	useReflect bool
 }
 
--- a/internal/codecs/array.go
+++ b/internal/codecs/array.go
@@ -42,20 +42,31 @@ func popArrayCodec(
 
 // Array is an EdgeDB array type codec.
 type Array struct {
-	id    types.UUID
-	child Codec
-	typ   reflect.Type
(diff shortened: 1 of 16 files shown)


fmoor:
This is based on https://github.com/edgedb/edgedb-go/pull/51. WIP until https://github.com/edgedb/edgedb-go/pull/51 is merged.

internal/codecs/array.go
```diff
@@ -42,20 +42,31 @@ func popArrayCodec(
...
 type Array struct {
-	id    types.UUID
-	child Codec
-	typ   reflect.Type
-	step  int
+	id         types.UUID
+	child      Codec
+	typ        reflect.Type
+	step       int
+	useReflect bool
```
1st1, line 49: Can you add some comments? It's unclear what `step` is or why `useReflect` is necessary

1st1 (commented):
overall lgtm but it's unclear to me what step & useReflect do without deep code reading

internal/codecs/array.go
fmoor: added :heavy_check_mark:

internal/codecs/array.go
```diff
@@ -45,17 +45,33 @@ type Array struct {
 	id    types.UUID
 	child Codec
 	typ   reflect.Type
-	step  int
+
+	// step is the element width in bytes for a go array of type `Array.typ`.
+	step int
```
1st1, line 50: Why is this needed btw? I don't recall having this field in Python or JS implementations.

fmoor: This is needed because decoding is done by writing data to known memory locations based on the out value's type. When there is a slice (an array under the hood) in the data structure the codec needs to know where each array element starts in memory. Something like:

```go
var head unsafe.Pointer

for i := 0; i < elemCount; i++ {
    decodeNextElementToLocation(head + i*step)
}
```

Issue #32, opened 2020-12-01
implement RFC1004


fmoor:
https://github.com/edgedb/rfcs/blob/master/text/1004-transactions-api.rst#edgedb-changes

PR #61, opened 2021-01-13
add reconnect on retryable errors

fmoor:
this feature is part of [RFC1004](https://github.com/edgedb/edgedb-go/issues/32)

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -16,9 +16,45 @@
 
 package edgedb
 
+import (
+	"context"
+
+	"github.com/edgedb/edgedb-go/internal/cache"
+)
+
 // Conn is a conn created outside of a pool.
 type Conn struct {
 	baseConn
+}
+
+// ConnectOne establishes a connection to an EdgeDB server.
+func ConnectOne(ctx context.Context, opts Options) (*Conn, error) { // nolint:gocritic,lll
+	return ConnectOneDSN(ctx, "", opts)
+}
+
+// ConnectOneDSN establishes a connection to an EdgeDB server.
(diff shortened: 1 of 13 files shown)


1st1:
This isn't yet ready for a review, right?

fmoor:
> This isn't yet ready for a review, right?

No not yet. I'll add you as a reviewer when it is ready.

conn.go
```diff
@@ -16,11 +16,47 @@
...
+	"github.com/edgedb/edgedb-go/internal/cache"
+)
+
 // Conn is a conn created outside of a pool.
 type Conn struct {
 	baseConn
 }
 
+// ConnectOne establishes a connection to an EdgeDB server.
+func ConnectOne(ctx context.Context, opts Options) (*Conn, error) { // nolint:gocritic,lll
```
1st1, line 31: I'm just curious what kind of linting rule does `nolint:gocritic,lll` disable?

1st1 (approved):

conn.go
fmoor: This repo is using a meta linter [golangci-lint](https://github.com/golangci/golangci-lint). The names are linters, [gocritic (general linter)](https://github.com/go-critic/go-critic) and [lll (line length linter)](https://github.com/walle/lll).

fmoor: Specifically, gocritic complains about passing `Options` by value (copies) rather than passing a pointer because `Options` is a large struct. However, I think that we want a copy here so that the user can't change the options later.

PR #62, opened 2021-01-15
add try transaction and retry

fmoor:
this is part of https://github.com/edgedb/edgedb-go/issues/32

- Interfaces are introduced to allow the most flexibility for future changes.
- Only the minimum transaction API is introduced to fulfill RFC1004 requirements. A fuller API if required is left for future work.
- Transaction is shortened to Tx because both [database/sql](https://golang.org/pkg/database/sql/#Tx) and [pgx](https://pkg.go.dev/github.com/jackc/pgx/v4#Tx) use this name and it seems to fit the Go naming paradigm of including the maximum information with the fewest characters.
- Interface names end in -er because that is the [go recommendation](https://golang.org/doc/effective_go.html#interface-names).

### DIFF
--- a/borrow.go
+++ b/borrow.go
@@ -16,32 +16,32 @@
 
 package edgedb
 
-import (
-	"errors"
-	"testing"
+type borrowable struct {
+	reason string
+}
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-)
+func (b *borrowable) assertUnborrowed() error {
+	switch b.reason {
+	case "transaction":
+		return &interfaceError{
+			msg: "Connection is borrowed for a transaction. " +
+				"Use the methods on transaction object instead.",
+		}
+	default:
+		return nil
+	}
+}
 
-func TestReleasePoolConn(t *testing.T) {
-	p := &pool{freeConns: make(chan *baseConn, 1)}
-	conn := &baseConn{}
-	pConn := &poolConn{pool: p, baseConn: conn}
+func (b *borrowable) borrow(reason string) error {
+	if b.reason != "" {
+		msg := "connection is already borrowed for " + b.reason
+		return &interfaceError{msg: msg}
+	}
 
-	err := pConn.Release()
-	require.Nil(t, err)
+	b.reason = reason
+	return nil
+}
 
-	result := <-p.freeConns
-	assert.Equal(t, conn, result)
-
-	err = pConn.Release()
-	assert.EqualError(
-		t,
-		err,
-		"edgedb.InterfaceError: connection released more than once",
-	)
-
-	var errReleasedTwice InterfaceError
-	assert.True(t, errors.As(err, &errReleasedTwice))
+func (b *borrowable) unborrow() {
+	b.reason = ""
 }
--- a/borrow.go
+++ b/borrow.go
@@ -16,6 +16,8 @@
 
 package edgedb
 
+import "fmt"
+
 type borrowable struct {
 	reason string
 }
@@ -27,8 +29,10 @@ func (b *borrowable) assertUnborrowed() error {
 			msg: "Connection is borrowed for a transaction. " +
 				"Use the methods on transaction object instead.",
 		}
-	default:
+	case "":
 		return nil
+	default:
+		panic(fmt.Sprintf("unexpected reason: %q", b.reason))
 	}
 }
 
@@ -38,10 +42,18 @@ func (b *borrowable) borrow(reason string) error {
 		return &interfaceError{msg: msg}
 	}
 
+	if reason != "transaction" {
+		panic(fmt.Sprintf("unexpected reason: %q", reason))
+	}
+
 	b.reason = reason
 	return nil
 }
 
 func (b *borrowable) unborrow() {
+	if b.reason == "" {
+		panic("not currently borrowed, can not unborrow")
+	}
+
 	b.reason = ""
 }
--- a/edgedb.go
+++ b/edgedb.go
@@ -59,8 +59,8 @@ type Trier interface {
 	Retry(context.Context, Action) error
 }
 
-// Querier allows querying the database.
-type Querier interface {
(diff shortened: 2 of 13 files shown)


tailhook (commented):
Looks good, except:
1. There are no tests on retry. Or am I missing something?
2. Conn -> baseConn split looks a little bit wrong. Currently reconnect works on `baseConn` level, so if I have something like this:
```
conn.TryTx(ctx, func(ctx context.Context, tx Tx) error {
    user := conn.QueryOne(ctx, "SELECT User")
    // connection is dropped here
    // now, next query will reconnect:
    conn.Execute(ctx, "UPDATE User SET { visits:= .visits + 1 }")
    // so last execute will happen outside of a transaction and commit will probably fail...
})
```
So in all other bindings, we don't reconnect inside equivalent of the `baseConn` itself, but rather in equivalent of `Conn`. But tie transaction to `baseConn`, so if connection crashed, subsequent requests on that `baseConn` just continue to fail, until someting higher level kicks in (i.e. next retry iteration, or new transaction from the `Conn` is started).

edgedb.go
```diff
@@ -33,8 +34,53 @@ import (
...
+	// otherwise it is committed.
+	TryTx(context.Context, Action) error
+
+	// Retry does the same as TryTx
+	// but retries the action if it might succeed on a subsequent attempt.
+	Retry(context.Context, Action) error
+}
+
+// Querier allows querying the database.
+type Querier interface {
```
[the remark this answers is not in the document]
fmoor: I used this name because it seemed to follow [this convention](https://golang.org/doc/effective_go.html#interface-names), but I'm happy to change it to `Executor`.

transaction_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func ensureTxTestType(t *testing.T) {
```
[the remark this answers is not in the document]
fmoor: >     1. There are no tests on retry. Or am I missing something?

There is a retry test in [`pool_test.go`](https://github.com/edgedb/edgedb-go/blob/d7c6d0add259be05248bc567844084119d64e91b/pool_test.go#L214)

fmoor:
>     1. There are no tests on retry. Or am I missing something?

there is a retry test in [pool_test.go](https://github.com/edgedb/edgedb-go/blob/d7c6d0add259be05248bc567844084119d64e91b/pool_test.go#L214)

fmoor:
> 2. Conn -> baseConn split looks a little bit wrong.

Makes sense. I'll fix this in a new PR since this one doesn't introduce reconnecting.  Issue here https://github.com/edgedb/edgedb-go/issues/67

transaction_test.go
```diff
@@ -0,0 +1,130 @@
...
+	ensureTxTestType(t)
+
+	ctx := context.Background()
+	err := conn.TryTx(ctx, func(ctx context.Context, tx Tx) error {
+		query := "INSERT TxTest {name := 'Test Roll Back'};"
+		if e := tx.Execute(ctx, query); e != nil {
+			return e
+		}
+
+		return tx.Execute(ctx, "SELECT 1 / 0;")
```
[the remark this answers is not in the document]
fmoor: added a test for rolling back on user defined errors

edgedb.go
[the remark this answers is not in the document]
fmoor: changed

transaction_test.go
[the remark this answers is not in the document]
fmoor: fixed

borrow.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (b *borrowable) assertUnborrowed() error {
+	switch b.reason {
+	case "transaction":
+		return &interfaceError{
+			msg: "Connection is borrowed for a transaction. " +
+				"Use the methods on transaction object instead.",
+		}
+	default:
+		return nil
```
1st1, line 31: We should panic if `reason` is some other string but `transaction` and return `nil` if it's `nil`.

borrow.go
```diff
@@ -0,0 +1,47 @@
...
+		msg := "connection is already borrowed for " + b.reason
+		return &interfaceError{msg: msg}
+	}
+
+	b.reason = reason
+	return nil
+}
+
+func (b *borrowable) unborrow() {
+	b.reason = ""
```
1st1, line 46: You perhaps should check if it was borrowed in the first place before "unborrowing"

borrow.go
fmoor: changed

borrow.go
fmoor: changed

edgedb.go
```diff
@@ -33,8 +34,53 @@ import (
...
+
+// Action is work to be done in a transaction.
+type Action func(context.Context, Tx) error
+
+// Trier allows trying actions in a transaction.
+type Trier interface {
+	// TryTx runs an action in a transaction.
+	// If the action returns an error the transaction is rolled back,
+	// otherwise it is committed.
+	TryTx(context.Context, Action) error
```
1st1, line 55: I've a question about the name. Why `TryTx` and not `TryTransaction` like in JS/Python drivers? Is `Tx` common in Go DB drivers?

fmoor: Transaction is shortened to Tx because both database/sql and pgx use this name and it seems to fit the Go naming paradigm of including the maximum information with the fewest characters.

1st1 (approved):
Awesome work. I'm not sure about the `TryTx` method name though.

1st1:
> Transaction is shortened to Tx because both database/sql and pgx use this name and it seems to fit the Go naming paradigm of including the maximum information with the fewest characters.

Alright, let's do that. Copying @elprans just so that he's aware (and has a chance to veto :))

elprans:
+1 to following the nomenclature established in the ecosystem.

Issue #67, opened 2021-01-19
don't reconnect while a Tx has borrowed a connection


fmoor:
see https://github.com/edgedb/edgedb-go/pull/62#pullrequestreview-571199597

Currently reconnecting can happen during an `Action` callback. This can result in only part of the action being committed. There should not be any reconnect attempts made during an `Action` callback.

```go
conn.TryTx(ctx, func(ctx context.Context, tx Tx) error {
    user := conn.QueryOne(ctx, "SELECT User")
    // connection is dropped here
    // now, next query will reconnect:
    conn.Execute(ctx, "UPDATE User SET { visits:= .visits + 1 }")
    // so last execute will happen outside of a transaction and commit will probably fail...
})
```

PR #68, opened 2021-01-19
prevent reconnecting during a transaction Action

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/67  
fixes https://github.com/edgedb/edgedb-go/issues/32

Previously the client would try to reconnect during an Action which could lead to failed or partial commits. This change makes it so the client doesn't reconnect during an Action. Instead it only checks the connection and reconnects if required before each Action attempt.


fmoor:
Based on https://github.com/edgedb/edgedb-go/pull/62. WIP until upstream is merged.

1st1:
> Previously the client would try to reconnect during an Action which could lead to failed or partial commits. This change makes it so the client doesn't reconnect during an Action. Instead it only checks the connection and reconnects if required before each Action attempt.

What happens if there's a disconnect during `Retry()` call?

fmoor:
> What happens if there's a disconnect during Retry() call?

`Retry()` ensures a connection before every retry attempt and only retries errors tagged with `SHOULD_RETRY`.  It looks like all connection errors tagged with `SHOULD_RETRY` are also tagged with `SHOULD_RECONNECT`. So to answer the question it would try to reconnect, then if successful re-run the callback.

1st1 (approved):

1st1:
> Retry() ensures a connection before every retry attempt and only retries errors tagged with SHOULD_RETRY. It looks like all connection errors tagged with SHOULD_RETRY are also tagged with SHOULD_RECONNECT. So to answer the question it would try to reconnect, then if successful re-run the callback.

Yeah, so the behavior is the same as in Python/JS; this answers my question.

Issue #64, opened 2021-01-18
documentaion review


fmoor:
- The documentation needs to be reviewed for completeness. 
- All `...` should be replaced with meaningful descriptions. 
- Examples should be added where appropriate.

fmoor:
After adding all the auto generated errors the godoc output is almost unusable. It is quite difficult to find anything or determine what is important in the enormous list of error names (see first image). It seems much more useful with all the errors removed (see second image). If the error types were moved to a separate package godoc would put the errors on their own page. I'm not sure if this is a good idea since the API is probably nicer to use with only one package. @1st1 what do you think?

![image](https://user-images.githubusercontent.com/8278635/104962823-79e5c580-5996-11eb-912a-f107e70d8487.png)

![image](https://user-images.githubusercontent.com/8278635/104963109-0a240a80-5997-11eb-9ced-402da0ab6f9f.png)

1st1:
Ю I'm not sure if this is a good idea since the API is probably nicer to use with only one package. @1st1 what do you think?

I think it's a good idea. Do you have any good example of how the separate package for errors would make code look worse?  We can probably keep the root `Error` (which should be called `EdgeDBError`, no?) type in the main package.

fmoor:
> the root Error (which should be called EdgeDBError, no?)

I don't think so. In go you must always use the package name when referring to imported identifiers, so this will always be `edgedb.Error`.

fmoor:
> Do you have any good example of how the separate package for errors would make code look worse?

In go you can't nest package names like this `edgedb.errors.SomeError` you can only write something like `errors.SomeError` so I'm not sure what the package should be called. There are already several prominent packages called `errors` and `error` is a built in type name. Maybe `edberr` or `edgedberr`. `edgedberr` feels long to me especially considering how long most of the error names are already. Some other name... My hesitation in breaking out a separate error package is only because `edgedb.SomeError` seems like the nicest name.

fmoor:
> We can probably keep the root Error type in the main package.

We could but why is it special? Keeping it in the main package might make be misleading for someone that is just skimming the documentation. It would look like there is only one error type. It seems better to me to keep all error types and `ErrorTag` in the same package whichever that ends up being.

1st1:
> We could but why is it special? Keeping it in the main package might be misleading for someone that is just skimming the documentation. It would look like there is only one error type. It seems better to me to keep all error types and ErrorTag in the same package whichever that ends up being.

[...]

PR #72, opened 2021-01-20
update documentation

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/64

To se a sample of the generated documentation checkout the repo and run `godoc` in the root directory. Then navigate to http://localhost:6060/pkg/github.com/edgedb/edgedb-go/. Note that the `godoc` tool generates subtly different output than `pkg.go.dev`. One major difference is that the later includes the repo's README.


fmoor:
This is WIP until the details for https://github.com/edgedb/edgedb-go/issues/64 are worked out. Specifically, if errors should be moved into their own package.

1st1 (approved):

Issue #44, opened 2020-12-19
assert data length in codecs


fmoor:
see https://github.com/edgedb/edgedb-go/pull/37#discussion_r546173288 for details

PR #77, opened 2021-01-25
assert data length

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/44

This change adds buff.Reader.PopSlice(n) which panics if the number of
bytes requested is more than the number of bytes available.

### DIFF
--- a/internal/codecs/named_tuple.go
+++ b/internal/codecs/named_tuple.go
@@ -145,39 +145,57 @@ func (c *NamedTuple) DecodeReflect(r *buff.Reader, out reflect.Value) {
 }
 
 func (c *NamedTuple) decodeReflectStruct(r *buff.Reader, out reflect.Value) {
-	r.Discard(4) // data length
 	elmCount := int(int32(r.PopUint32()))
 
 	for i := 0; i < elmCount; i++ {
 		r.Discard(4) // reserved
+
+		elmLen := r.PopUint32()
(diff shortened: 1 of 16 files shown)


fmoor:

internal/codecs/named_tuple.go
```diff
@@ -145,39 +145,57 @@ func (c *NamedTuple) DecodeReflect(r *buff.Reader, out reflect.Value) {
...
 
 // DecodePtr decodes a named tuple into an unsafe.Pointer.
 func (c *NamedTuple) DecodePtr(r *buff.Reader, out unsafe.Pointer) {
-	r.Discard(4) // data length
 	elmCount := int(int32(r.PopUint32()))
 
 	for i := 0; i < elmCount; i++ {
 		r.Discard(4) // reserved
+
+		elmLen := int(int32(r.PopUint32()))
```
1st1, line 192: I'm just curious, why not keep it int32?

1st1 (approved):

internal/codecs/named_tuple.go
fmoor: Because PopSlice() takes int, but I guess PopSlice is always used with a value decoded from a uint32 field :thinking:. I'll change PopSlice() to take uint32.

fmoor: changed

Issue #39, opened 2020-12-18
handle unexpected message in background correctly


fmoor:
per https://github.com/edgedb/edgedb-go/pull/37#discussion_r545489144

When the client is not being used i.e. it is listening for server initiated messages in the background and an unexpected message is received, the client should close it's connection to the server and return an error to the user the next time it is used.

PR #79, opened 2021-01-26
close connection on unexpected message

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/39

### DIFF
--- a/edgedb.go
+++ b/edgedb.go
@@ -116,7 +116,8 @@ type Executor interface {
 }
 
 type baseConn struct {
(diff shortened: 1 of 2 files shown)


fmoor:

1st1 (approved):
Yeah, at some point we'll need to figure out how we can test things like these across all our driver implementations.  Perhaps invest into an EdgeDB proxy server that would be used in tests to inject bad messages into the protocol and see how our drivers react to that.

Issue #5, opened 2020-11-04
improve out type error messages


fmoor:
https://github.com/edgedb/edgedb-go/pull/1#discussion_r517563293

PR #82, opened 2021-01-26
improve type miss-match error messages

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/5

This change adds the data structure field path to type miss-match error messages. An example error message would look like:
`expected mypackage.MyType.path.to.field to be int64, got int`


fmoor:

1st1 (approved):
Great DX improvement!

PR #86, opened 2021-02-02
fix uuid encoding

fmoor:
This change fixes a bug that made it impossible to use UUIDs as query arguments.

### DIFF
--- a/internal/buff/read.go
+++ b/internal/buff/read.go
@@ -20,8 +20,8 @@ import (
 	"encoding/binary"
 	"fmt"
 
+	types "github.com/edgedb/edgedb-go/internal/edgedbtypes"
 	"github.com/edgedb/edgedb-go/internal/soc"
-	"github.com/edgedb/edgedb-go/internal/types"
 )
 
 // Reader is a buffer reader.
--- a/internal/buff/read_test.go
+++ b/internal/buff/read_test.go
@@ -19,8 +19,8 @@ package buff
 import (
 	"testing"
 
+	types "github.com/edgedb/edgedb-go/internal/edgedbtypes"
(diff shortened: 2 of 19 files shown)


1st1 (approved):
LGTM.

Question: say I'm creating a web app. I receive a UUID via HTTP GET request in its text form (like `"00010203-0405-0607-0809-0a0b0c0d0e0f"`). How can I pass that as a query argument to EdgeDB?

fmoor:
> say I'm creating a web app. I receive a UUID via HTTP GET request in its text form (like `"00010203-0405-0607-0809-0a0b0c0d0e0f"`). How can I pass that as a query argument to EdgeDB?

This is the current API. MarshalText and UnmarshalText allow edgedb.UUID to [play nice](https://golang.org/pkg/encoding/#TextMarshaler) with json and other encodings in the standard library. 
```go
str := "00010203-0405-0607-0809-0a0b0c0d0e0f"

var id edgedb.UUID
err := id.UnmarshalText([]byte(str))
if err != nil {
    log.Fatal(err)
}

con.QueryOne(ctx, "SELECT User WHERE .id = <uuid>$0", &result, id)
```

1st1:
> This is the current API. MarshalText and UnmarshalText allow edgedb.UUID to play nice with json and other encodings in the standard library.

What's the downside of allowing UUID to encode/decode to/from Go `string` type directly (in addition to `edgedb.UUID`)?

fmoor:
> What's the downside of allowing UUID to encode/decode to/from Go string type directly (in addition to edgedb.UUID)?

I don't have any objections. Google has that in [their implementation](https://pkg.go.dev/github.com/google/uuid#Parse).

PR #89, opened 2021-02-03
add cal::local_datetime codec

fmoor:
working on https://github.com/edgedb/edgedb-go/issues/48

### DIFF
--- a/internal/edgedbtypes/datetime.go
+++ b/internal/edgedbtypes/datetime.go
@@ -1,0 +1,50 @@
+// This source file is part of the EdgeDB open source project.
+//
+// Copyright 2020-present EdgeDB Inc. and the EdgeDB authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package edgedbtypes
+
+import (
+	"time"
+)
+
+// timeShift is the number of microseconds
+// between 0001-01-01T00:00 and 2000-01-01T00:00
+const timeShift = 62_135_596_800_000_000
+
+// NewLocalDateTime returns a new LocalDateTime
+func NewLocalDateTime(
+	year int, month time.Month, day, hour, minute, second, microsecond int,
+) LocalDateTime {
+	t := time.Date(
+		year, month, day, hour, minute, second, microsecond*1_000, time.UTC,
+	)
+	sec := t.Unix()
+	nsec := int64(t.Sub(time.Unix(sec, 0)))
+	return LocalDateTime{sec*1_000_000 + nsec/1_000 + timeShift}
+}
+
+// LocalDateTime is a date and time without timezone.
+// https://www.edgedb.com/docs/datamodel/scalars/datetime/
+type LocalDateTime struct {
+	usec int64
+}
+
+func (dt LocalDateTime) String() string {
+	usec := dt.usec - timeShift
+	sec := usec / 1_000_000
+	nsec := (usec % 1_000_000) * 1_000
+	return time.Unix(sec, nsec).UTC().Format("2006-01-02T15:04:05.999999")
+}
--- a/internal/codecs/base_scalar.go
+++ b/internal/codecs/base_scalar.go
@@ -22,7 +22,6 @@ import (
 	"math"
 	"math/big"
 	"reflect"
-	"time"
 	"unsafe"
 
 	"github.com/edgedb/edgedb-go/internal/buff"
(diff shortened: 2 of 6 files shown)


fmoor:
This change is based on https://github.com/edgedb/edgedb-go/pull/85 & https://github.com/edgedb/edgedb-go/pull/86, WIP until upstream changes are merged.

internal/codecs/datetime_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLocalDateTimeDecode(t *testing.T) {
+	r := buff.SimpleReader([]byte{
+		0x00, 0x02, 0x2b, 0x35, 0x9b, 0xc4, 0x10, 0x00,
+	})
+
+	var result types.LocalDateTime
```
1st1, line 34: For edgedb-go users it will be `edgedb.LocalDateTime`, right?

internal/codecs/datetime_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"unsafe"
+
+	"github.com/edgedb/edgedb-go/internal/buff"
+	types "github.com/edgedb/edgedb-go/internal/edgedbtypes"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLocalDateTimeDecode(t *testing.T) {
+	r := buff.SimpleReader([]byte{
```
1st1, line 30: BTW, I'm not sure if we really need low-level byte IO tests like this. Up to you, but wouldn't it be enough to just have high-level functional tests where we query the DB with these types?

1st1 (approved):

internal/codecs/datetime_test.go
fmoor: correct

internal/codecs/datetime_test.go
fmoor: Yes it would be enough to identify that something is wrong. I feel like tests like this can help identify what specifically is wrong. That said, I am open to getting rid of them. They do incure a certain amount of development over head.

1st1: > They do incure a certain amount of development over head.

Yeah, this is why I'm asking. We don't have tests like this in edgedb-js or edgedb-python. So if maintaining them is tedious I'd drop them. Also looking at `x00, 0x02, 0x2b, 0x35, 0x9b, 0xc4, 0x10, 0x00` doesn't show me a clear picture of the data layout, I'd still need to debug it to understand what went wrong.

internal/edgedbtypes/datetime.go
```diff
@@ -0,0 +1,47 @@
...
+	)
+	sec := t.Unix() - timeShift
+	nsec := int64(t.Sub(time.Unix(t.Unix(), 0)))
+	return LocalDateTime{sec*1_000_000 + nsec/1_000}
+}
+
+// LocalDateTime is a date and time without timezone.
+// https://www.edgedb.com/docs/datamodel/scalars/datetime/
+type LocalDateTime struct {
+	usec int64
```
fmoor, line 40: @1st1 @tailhook @elprans I updated this with the feedback from https://github.com/edgedb/edgedb-go/pull/90#discussion_r569736492 let me know if I am on the right track.

1st1: So this is now a proper unix timestamp?

fmoor: No it is still a 2000's time stamp but that is an implementation detail except for the zero value. :thinking: i guess `LocalDateTime{} == "1970-01-01T00:00"` makes more sense than `"2000-01-01T00:00"`. I can change that. What about the rest of the public API?

1st1: > No it is still a 2000's time stamp but that is an implementation detail except for the zero value. 

The zero value is what bothers me. I'd expect it to be `00/00/0000`, not some weird 2000s date.

Also the argument about `9999/9999/9999` kind of values is IMO invalid. The max possible sane LocalDate would be `9999/12/31`.

1st1: IOW I'm not so sure we should use the timestamp internally. This type, again, is more about wall calendar, than about all possible future/past timelines. @elprans @tailhook

elprans: I have no problem with using a timestamp internally as long as it is opaque, i.e. there should not be a (documented?) way of setting or getting the raw value from user code.  The timestamp approach has advantages: it's memory-efficient and makes comparisons easy and fast.

1st1: > I have no problem with using a timestamp internally as long as it is opaque, i.e. there should not be a (documented?) way of setting or getting the raw value from user code.

Well, that's the crux of the problem. It will be opaque, unless you just create an empty struct, like `edgedb.LocalDate{}`, in which case the local date you'd get wouldn't be `0/0/0`, it would be `2000/01/30` or something.

fmoor: > It will be opaque, unless you just create an empty struct, like edgedb.LocalDate{}, in which case the local date you'd get wouldn't be 0/0/0, it would be 2000/01/30 or something.

We can make the zero value be `0/0/0` no problem.

1st1: Let's keep it simple as Elvis & Paul suggested, with one field in the struct.

1st1 (approved):

PR #90, opened 2021-02-03
add cal::local_date codec

fmoor:
working on https://github.com/edgedb/edgedb-go/issues/48

### DIFF
--- a/internal/edgedbtypes/datetime.go
+++ b/internal/edgedbtypes/datetime.go
@@ -22,9 +22,11 @@ import (
 	"time"
 )
 
-// timeShift is the number of microseconds
-// between 0001-01-01T00:00 and 2000-01-01T00:00
-const timeShift = 62_135_596_800_000_000
+const (
+	// timeShift is the number of seconds
+	// between 0001-01-01T00:00 and 1970-01-01T00:00
+	timeShift = 62135596800
+)
 
 // NewLocalDateTime returns a new LocalDateTime
 func NewLocalDateTime(
@@ -33,9 +35,9 @@ func NewLocalDateTime(
 	t := time.Date(
 		year, month, day, hour, minute, second, microsecond*1_000, time.UTC,
 	)
-	sec := t.Unix()
-	nsec := int64(t.Sub(time.Unix(sec, 0)))
-	return LocalDateTime{sec*1_000_000 + nsec/1_000 + timeShift}
+	sec := t.Unix() + timeShift
+	nsec := int64(t.Sub(time.Unix(t.Unix(), 0)))
+	return LocalDateTime{sec*1_000_000 + nsec/1_000}
 }
 
 // LocalDateTime is a date and time without timezone.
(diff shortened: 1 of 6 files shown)


fmoor:
This change is based on https://github.com/edgedb/edgedb-go/pull/89. WIP until upstream change is merged.

internal/edgedbtypes/datetime.go
```diff
@@ -16,8 +16,17 @@
...
+// LocalDateTime is a date and time without timezone
+// represented as the number of microseconds between the encoded datetime
+// and January 1st 2000, 00:00 UTC
 // https://www.edgedb.com/docs/datamodel/scalars/datetime/
 type LocalDateTime struct {
 	Microseconds int64
 }
+
+// LocalDate is a date without a time zone
+// represented as days between the encoded date and January 1st 2000
```
1st1, line 28: That's the internal encoding of it (postgres has this weird timestamp starting from 2000s). I'm not sure if that should leak into user code. @elprans @tailhook

fmoor: Are you thinking it should be number of days since the epoch `1970-01-01`?

1st1: That would make it less weird, for sure. But since it's LocalDate, I'm wondering if it's possible to just literally decode '2020/11/17' into `{year: 2020, month: 11, day: 17}`.

fmoor: If you want to have those fields and make them public, what are you going to do when someone uses `{year: 9999, month: 9999, day: 9999}` as a query argument? It seems like there are more edge cases with that memory layout. If the motivation is to make the API more human friendly perhaps the field(s) should be private and provide methods for accessing conceptual things like `Day()` and `Month()`. Another option would be to make this (and other cal::local*) `type LocalDate int32`. This would enable comparison operators `<` & `>` and it can still have friendly methods.

tailhook: In Rust we always ensure that structure fields are consistent. Where value range is a part of the consistency guarantees. So hiding everything except plain "dataclasses" behind private fields and methods is a must.

In Go it's much more common to have rules that aren't enforced by the type system. The problem with hiding internal structure in Go, is that it isn't full encapsulation because you can create a zero-filled structure in cany case, so relying on any encapsulation in Go is moot idea. Anyway `type LocalDate int32` is too weak option. But whether we want to hide internal structure of `LocalDate` depends on what we have in all other structures, so the access pattern is clear.

I would go for hiding everything since that would mean more compact structures (i.e. 4 byte `days` instead >= 8 bytes for 3 fields) or allow us more robustness in case we eventually allow non-postgres edgedb backend.

fmoor: I made changes to LocalDateTime in an upstream branch https://github.com/edgedb/edgedb-go/pull/89#discussion_r570435295

fmoor: OK, going with with one field in the struct and a zero value of `1/1/1`. See https://github.com/edgedb/edgedb-go/pull/89#discussion_r570435295 for additional discussion.

Issue #88, opened 2021-02-03
correct duration codec type


fmoor:
Currently `std::duration` is decoded into go's `time.Duration`. However, `time.Duration` is int64 as nanoseconds where `std::duration` is int64 as microseconds. The go type should be changed to support the same scale and range as `std::duration`.

PR #93, opened 2021-02-05
add edgedb.Duration type

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/88

### DIFF
--- a/internal/edgedbtypes/datetime_test.go
+++ b/internal/edgedbtypes/datetime_test.go
@@ -69,3 +69,71 @@ func TestNewLocalDateTime(t *testing.T) {
 		})
 	}
 }
+
+func TestDurationString(t *testing.T) {
+	samples := []struct {
+		str string
+		d   Duration
+	}{
+		{"1us", Duration(1)},
+		{"999us", Duration(999)},
+		{"-1us", Duration(-1)},
+		{"-999us", Duration(-999)},
+
+		{"1.234ms", Duration(1_234)},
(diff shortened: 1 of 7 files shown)


internal/edgedbtypes/datetime_test.go
```diff
@@ -69,3 +69,71 @@ func TestNewLocalDateTime(t *testing.T) {
...
+		{"854015929h20m18.258432s", Duration(3074457345618258432)},
+		{"-59m59s", Duration(-3599000000)},
+		{"-1h59m59s", Duration(-7199000000)},
+		{"-1h59m", Duration(-7140000000)},
+		{"-854015929h20m18.258432s", Duration(-3074457345618258432)},
+	}
+
+	for _, s := range samples {
+		t.Run(s.str, func(t *testing.T) {
+			assert.Equal(t, s.str, s.d.String())
```
1st1, line 136: Can you also add a functional test with EdgeDB, where you send duration converted from a string like "-1h35m17s" and convert it to string in EdgeQL, and then compare the result on the client side?

1st1 (approved):

internal/edgedbtypes/datetime_test.go
fmoor: added :heavy_check_mark: 

I  had to change `μs` to `us` because of https://github.com/edgedb/edgedb/issues/2194 and I removed a test case because of https://github.com/edgedb/edgedb/issues/2195

Issue #96, opened 2021-02-09
prevent ad-hock transactions


fmoor:
Currently if a connection is acquired either by `ConnectOne()` or `pool.Acquire()` and a transaction is started using `START TRANSACTION` instead of TryTx() or Retry() the connection might reconnect mid transaction.

```go
con.Execute(ctx, "START TRANSACTION")
con.Query(...) // connections fails
con.Query(...) // new connection i.e. different transaction
con.Execute(ctx, "COMMIT")
```

Queries outside of TryTx() or Retry() should not allow explicit transactions.

tailhook:
I would prefer disallowing `START TRANSACTION`, rather than making hacks around transaction state. This is what `ALLOW_CAPABILITIES` is for. Is there any reason somebody needs `START TRANSACTION` instead of `TryTx`?

fmoor:
> I would prefer disallowing START TRANSACTION, rather than making hacks around transaction state.

+1

fmoor:
updated description & title

PR #98, opened 2021-02-09
disallow ad-hock transactions in pool queries

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/97

PR #99, opened 2021-02-10
refuse transactions outside of TryTx & Retry

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/96

### DIFF
--- a/internal/header/header.go
+++ b/internal/header/header.go
@@ -1,0 +1,37 @@
+// This source file is part of the EdgeDB open source project.
+//
+// Copyright 2020-present EdgeDB Inc. and the EdgeDB authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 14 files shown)


fmoor:
Based on https://github.com/edgedb/edgedb-go/pull/98. WIP until upstream is merged.

internal/header/header.go
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package header
+
+import "encoding/binary"
+
+const (
+	// AllowCapabilities tells the server what capabilities it should allow.
+	AllowCapabilities        = 0xFF04
```
1st1, line 23:
```suggestion
	AllowCapabilitiesHeader        = 0xFF04
```

1st1 (approved):

internal/header/header.go
fmoor: The module name `header` will always be present when this is used in the edgedb package. Adding `Header` is redundant.

Issue #63, opened 2021-01-18
update README.md before beta release


fmoor:
Make sure that the example and other details in the README are correct before announcing the first beta release.

PR #100, opened 2021-02-10
update README

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/63

### DIFF
--- a/generatederrors.go
+++ b/generatederrors.go
@@ -22,90 +22,90 @@ package edgedb
 import "fmt"
 
 const (
+	ShouldRetry     ErrorTag = "SHOULD_RETRY"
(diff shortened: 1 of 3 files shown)


generatederrors.go
```diff
@@ -22,90 +22,90 @@ package edgedb
 import "fmt"
 
 const (
+	ShouldRetry     ErrorTag = "SHOULD_RETRY"
```
1st1, line 25: Are these changes related to the PR?

fmoor: Yes. The README says you can format the code using `make format` but that file was never formatted. I figured it would be strange for someone to run `make format` in a fresh checkout and get a huge diff. I also added `make format` to the error generation so that happens automatically in the future.

Issue #27, opened 2020-11-24
functional test review


fmoor:
The tests in this project that run against a real server should be compared with the tests in https://github.com/edgedb/edgedb-js and https://github.com/edgedb/edgedb-python to be sure that there are not any missed cases.

PR #103, opened 2021-02-17
fix Duration.String() bug

fmoor:
Found a bug in Duration.String() formatting while working on https://github.com/edgedb/edgedb-go/issues/27.


connutils_test.go
```diff
@@ -437,7 +437,7 @@ func TestConnectTimeout(t *testing.T) {
 		User:               opts.User,
 		Password:           opts.Password,
 		Database:           opts.Database,
-		ConnectTimeout:     1 * time.Millisecond,
+		ConnectTimeout:     2 * time.Nanosecond,
```
1st1, line 440: Why 2ns and not 1ns? :)

fmoor: The ConnectTimeout should be longer than WaitUntilAvailable so that the timeout is caused by the ConnectTimeout duration and not the WaitUntilAvailable duration.

1st1 (approved):

PR #113, opened 2021-02-26
add edgedb.ParseUUID

fmoor:
Make it easy to get a UUID from a string.

### DIFF
--- a/internal/edgedbtypes/uuid.go
+++ b/internal/edgedbtypes/uuid.go
@@ -23,6 +23,27 @@ import (
 	"strings"
 )
 
+// ParseUUID parses s into a UUID or returns an error.
+func ParseUUID(s string) (UUID, error) {
+	s = strings.ReplaceAll(s, "-", "")
+	if len(s) != 32 {
+		return UUID{}, errMalformedUUID
+	}
+
+	var tmp UUID
+	for i := 0; i < 16; i++ {
+		val, err := strconv.ParseUint(s[:2], 16, 8)
+		if err != nil {
+			return UUID{}, errMalformedUUID
+		}
+
+		tmp[i] = uint8(val)
+		s = s[2:]
+	}
+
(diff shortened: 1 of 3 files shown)


internal/edgedbtypes/uuid.go
```diff
@@ -22,6 +22,24 @@ import (
 	"strings"
 )
 
+// ParseUUID parses s into a UUID or returns an error.
+func ParseUUID(s string) (UUID, error) {
+	s = strings.Replace(s, "-", "", 4)
```
1st1, line 27: I'd add validation of string length here -- must be 32.

internal/edgedbtypes/uuid.go
```diff
@@ -22,6 +22,24 @@ import (
...
+
+	var id UUID
+	for i := 0; i < 16; i++ {
+		val, err := strconv.ParseUint(s[:2], 16, 8)
```
1st1, line 31: Perhaps a faster way would be to manually unroll this and parse it as two int64. You can easily break them into individual bytes after.

1st1 (approved):

internal/edgedbtypes/uuid.go
fmoor: If we are looking for optimal performance maybe we should do something like [this](https://github.com/google/uuid/blob/bfb86fa49a73e4194d93bea18d7acfe3694438ce/uuid.go#L84)?

internal/edgedbtypes/uuid.go
fmoor: changed

internal/edgedbtypes/uuid.go
1st1: Oh wow, they went to great length to avoid copying that string.  Up to you, actually.  I'm fine even if you merge this PR as is.

PR #115, opened 2021-03-02
fix panic when decoding malformed UUID

fmoor:


internal/edgedbtypes/uuid.go
```diff
@@ -45,6 +46,10 @@ func (id UUID) MarshalText() ([]byte, error) {
 // UnmarshalText unmarshals the id from a string.
 func (id *UUID) UnmarshalText(b []byte) error {
 	s := string(b)
```
1st1, line 48: `s` is always coming from the DB? How can it be malformed?

fmoor: This is for decoding JSON. Who knows where it is coming from. For example, decoding the json body of a post request that you will then use as arguments to an insert.

fmoor: This is never used inside an edgedb query method.

1st1: I still don't follow. Can you come up with an example?

fmoor: You get a POST request with a json payload. One or more of the values in that payload are UUIDs that you want to then use in a query. The json package uses the [TextUnmarshaler interfaces](https://golang.org/pkg/encoding/#TextUnmarshaler) to decode to types it doesn't know about. 

```go
type Movie struct {
    Title string
    Director struct {
        ID edgedb.UUID
    }
}

// CreateMovie is a request handler
func CreateMovie(c *gin.Context) {
    var payload Movie
    err := json.NewDecoder(c.Request.Body).Decode(&payload)
    if err != err { 
        panic(err) 
    }

    err = con.QueryOne(c, `
        INSERT Movie {
            title := <str>$0,
            director := (SELECT Director FILTER .id = <uuid>$1 LIMIT 1),
        }`,
        &result,
        payload.Title,
        payload.Director.ID,
    )

    ...
}
```

1st1: Ah, got it, thanks for explaining!

internal/edgedbtypes/uuid.go
```diff
@@ -42,16 +43,25 @@ func (id UUID) MarshalText() ([]byte, error) {
 	return []byte(id.String()), nil
 }
 
+var errMalformedUUID = errors.New("malformed edgedb.UUID")
+
 // UnmarshalText unmarshals the id from a string.
 func (id *UUID) UnmarshalText(b []byte) error {
 	s := string(b)
+	if len(s) != 36 {
```
1st1, line 51: I'd suggest different logic here:

`if len(s) < 32 || len(s) > 36 { return errMalformedUUID }`

This is what we have in EdgeDB: https://github.com/MagicStack/py-pgproto/blob/90ad4154a0aa6923e2f97d3d3d779c4b830ff415/uuid.pyx#L62-L75

The idea is that uuid string serilization is kinda arbitrary with respect to where the dashes are. Ultimately uuids are 128bit integers, and dashes are there to make them more readable to humans. E.g. `B7B69960-26D1-4BD4-8182876A1CC38BEE` can be just as valid as `B7B69960-26D1-4BD4-8182-876A1CC38BEE`.

Python also uses the same approach in its stdlib: https://github.com/python/cpython/blob/04f6fbb6969e9860783b9ab4dc24b6fe3c6dab8d/Lib/uuid.py#L173-L178

1st1: That said I don't think we need to handle `urn:` and `uuid:` prefixes just yet.

fmoor: changed to remove all `-`s and then assert `len(32)`

internal/edgedbtypes/uuid_test.go
```diff
@@ -47,3 +47,20 @@ func TestUUIDUnmarshalJSON(t *testing.T) {
 	expected := UUID{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}
 	assert.Equal(t, expected, uuid)
 }
+
+func TestUUIDUnmarshalJSONInvalid(t *testing.T) {
+	samples := []string{
+		`""`,
+		`"000102030405060708090a0b0c0d0e"`,
+		`"00010203-0405-060700809-0a0b0c0d0e0f"`,
```
1st1, line 55:
```suggestion
		`"00010203-0405-060700809-0a0b0c0d0e0f"`,
		`"00010203-0405-06070-08090a0b0c-0d0e0f"`,		
```

1st1 (approved):

Issue #114, opened 2021-03-01
remove __tid__ field skipping


fmoor:
It may be possible to have the server not send the __tid__ field. If possible have the server not send this field and remove all the code for skipping it in codecs.

Issue #47, opened 2020-12-28
implement decimal codec


fmoor:

fmoor:
I'm not sure what the right go type is for this codec. The options seems to be:

1. [big.Rat](https://golang.org/pkg/math/big/#Rat) (not fixed-point decimal precision)
2. [big.Float](https://golang.org/pkg/math/big/#Float) (not fixed-point decimal precision)
3. [github.com/shopspring/decimal](https://github.com/shopspring/decimal) (3rd party) - seems to be maintained and widely used, but has [limitations](https://github.com/shopspring/decimal/issues/207).
4. [github.com/cockroachdb/apd](https://github.com/cockroachdb/apd/v2) (3rd party) - maintained, full featured, more performant than `3`
5. Our own type that knows how to convert it's value to some or all of the above?

We can support more than one target type so the options are not mutually exclusive.

1st1:
I don't want us to bundle or adapt to some specific 3rd party library. Ideally we need to figure out an API that would allow users to plug their library of choice. Take a look at how asyncpg handles custom codecs. Go is of course a different beast, but maybe we can come up with a good solution.

fmoor:
I think something similar to go's [marshaler interfaces](https://golang.org/pkg/encoding/#BinaryMarshaler) would work well for this. This interface would only be used for EdgeDB scalar values excluding arrays.

```go
package edgedb

type Marshaler interface  {
    MarshalEdgeDB() ([]byte, error)
}

type Unmarshaler interface {
    UnmarshalEdgeDB([]byte) error
}
```

Instead of registering a codec for the edgedb type, go values that implement the interface will marshal themselves. This has a few advantages.

1. this pattern is familiar to go developers
2. you don't have to configure the client with special codecs

4thel00z:
Greetings @1st1 I am not sure why there is this DIY approach, if there is proper stdlib primitive that can be used as sensible defaults. Decimal is usually used for prices which are floats.
This is probably like 90% of all use cases. Can big.Rat (no pun intended) be used as default ?
Happy to contribute a PR in case of a yes.

fmoor:
> Can big.Rat (no pun intended) be used as default?

[...]

PR #119, opened 2021-03-11
add support for user defined encoding & decoding

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/47
fixes https://github.com/edgedb/edgedb-go/issues/114

The following changes were made:

- create a distinct descriptor type. Reading a descriptor off the wire goes form leaves to root of the type tree. Applying a go type to a descriptor to create a codec goes from root to leaves. Having a distinct descriptor type makes it easier to move root to leaves through both type trees at the same time to build a codec.
- split codecs into Encoders and Decoders to enable separate Marshaler and Unmarshaler interfaces.
- move scalar codecs to smaller files to keep file sizes reasonable.
- remove logic for skipping \_\_tid__ fields
- add support for Marshler and Unmarshaler interfaces which allow custom encoding and decoding for all scalar types except arrays. This API was inspired by the stdlib https://golang.org/pkg/encoding/#BinaryMarshaler
- test Marshaler and Unmarshaler for all scalar types.

# Custom Codec Example
```go
type DecimalUnmarshaler struct {
    // a real implementation would probably do more than store the raw bytes off the wire
    data []byte
}

func (d *DecimalUnmarshaler) UnmarshalEdgeDBDecimal(data []byte) error {
    // decode data from the server
    d.data = make([]byte, len(data))
    copy(d.data, data)
    return nil
}

func main() {
    var result DecimalMarshaler
    err := con.Query(ctx, "SELECT 1234n", &result)
    if err != nil {
        log.Fatal(err)
    }

    fmt.Println(result)
}
```


types.go
```diff
@@ -33,6 +36,110 @@ type (
...
+	BoolMarshaler = codecs.BoolMarshaler
+
+	// BoolUnmarshaler enables decoding user defined bool values.
+	BoolUnmarshaler = codecs.BoolUnmarshaler
+
+	// BytesMarshaler enables encoding user defined bytes values.
+	BytesMarshaler = codecs.BytesMarshaler
+
+	// BytesUnmarshaler enables decoding user defined bytes values.
+	BytesUnmarshaler = codecs.BytesUnmarshaler
```
fmoor, line 50: This long list of types clutters the documentation for the edgedb package. These types don't need to be public. User code doesn't need access to them. They just need to be documented. It is probably better either:

1. to leave them in `internal/codecs`, don't make them part of the public API, and reference them in the introduction to the edgedb package telling users that they can implement them.
2. create a second public package that only has these interfaces. Maybe `edgedbmarshalers` or something. This way it is obvious that they are part of the public API but it keeps the documentation cleaner.

1st1: > to leave them in internal/codecs, don't make them part of the public API, and reference them in the introduction to the edgedb package telling users that they can implement them.

I'd go with this option.

1st1:
Is there any performance degradation caused by the new, more advanced, marshal/unmarshal code?

1st1:
The type name in this example is a bit confusing:

```go
type DecimalUnmarshaler struct {
    // a real implementation would probably do more than store the raw bytes off the wire
    data []byte
}
```

IMO it would be more readable if the struct was named `CustomDecimal`. Same applies to unittests/docs.

1st1 (approved):
Solid work and everything looks good and well thought out. Before merging this though, I'm wondering what's the impact on performance looks like. We now switch based on `val.(type)` in every marshaller.

fmoor:
> I'm wondering what's the impact on performance looks like. We now switch based on `val.(type)` in every marshaller.

I'll run the benchmarks on it. I would be if this change made a significant difference. That type switch used to be a type assertion which is a very similar mechanism.

types.go
fmoor: changed

fmoor:
It doesn't seem like there is a change in performance with this PR.

# Before
image

# After
image

Issue #116, opened 2021-03-05
improve syntax error messages


fmoor:
Currently errors caused by invalid syntax only have a terse message about what is wrong, but don't have any context like where in the query the error occurred (line/column). When an error is returned because of a malformed query the error text should point out where in the schema the error is similar to the cli's error messages.

PR #120, opened 2021-03-12
add context to syntax errors

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/116

A query snip-it with arrow and hint is introduced for syntax errors.

## Caveats 
This implementation is very crude and likely to work poorly with characters outside of the ascii range because bytes are construed as characters. It also replaces tabs with spaces in the query to avoid misalignment when printed.

There doesn't seem to be a library in go similar to rust's [codespan](https://github.com/brendanzab/codespan) so the only option seems to be to roll our own.

## Example
`SELECT (foo (((1 2) 3)) 4)` produces
```
edgedb.EdgeQLSyntaxError: Unexpected token: <Token ICONST "2">
query:1:18

SELECT (foo (((1 2) 3)) 4)
                 ^ It appears that a ',' is missing in a tuple before '2'
```

### DIFF
--- a/error.go
+++ b/error.go
@@ -18,6 +18,10 @@ package edgedb
 
 import (
 	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"unicode/utf8"
 
 	"github.com/edgedb/edgedb-go/internal/buff"
 )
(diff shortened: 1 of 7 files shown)


fmoor:
If it seems worth while I could extend this to work with arbitrary unicode instead of just the ascii range.

fmoor:
This change is based on https://github.com/edgedb/edgedb-go/pull/119 and is WIP until the upstream branch is merged.

1st1:
> If it seems worth while I could extend this to work with arbitrary unicode instead of just the ascii range.

Yes, we should do that. The updated error messages is a major improvement.

1st1:
I'll wait until the review until https://github.com/edgedb/edgedb-go/pull/119 is merged.

error.go
```diff
@@ -51,19 +54,64 @@ func firstError(a, b error) error {
...
+	headers := make(map[uint16]string, n)
 	for i := 0; i < n; i++ {
-		r.PopUint16() // key
-		r.PopString() // value
+		headers[r.PopUint16()] = r.PopString()
 	}
 
-	return err
+	// if the error is syntax related
+	if 0x04_01_00_00 <= code && code < 0x04_02_00_00 {
```
1st1, line 86: Well, all kinds of errors can have position information in them. IMO we should instead check for that, and then render a nice message for all kinds of errors. E.g. a query might be correct syntactically, but have an invalid type reference; the server would return the exact spot of the problem in the query.

fmoor: changed

fmoor:
> I could extend this to work with arbitrary unicode instead of just the ascii range.

This is a little better now, but it turns out that character width, even in a fixed width font, is font dependent. `世界` is 2 code points but each character takes up 2 columns in my terminal font. I haven't found a nice way of determining the number of columns a grapheme cluster typically consumes.

1st1 (approved):

Issue #122, opened 2021-03-15
RFC1004 transaction & retry options


fmoor:
add ability to configure transaction options and transaction retrying options

PR #123, opened 2021-03-18
add transaction and retry options (RFC1004)

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/122

- adds the following methods to Conn, Pool, and PoolConn
```go
  // WithTxOptions returns a shallow copy of the pool
  // with the TxOptions set to opts.
  func (p Pool) WithTxOptions(opts TxOptions) *Pool

  // WithRetryOptions returns a shallow copy of the pool
  // with the RetryOptions set to opts.
  func (p Pool) WithRetryOptions(opts RetryOptions) *Pool
```
- adds type TxOptions struct
```go
  // WithIsolation returns a copy of the TxOptions
  // with the isolation level set to i.
  func (o TxOptions) WithIsolation(i IsolationLevel) TxOptions

  // WithReadOnly returns a shallow copy of the pool
  // with the transaction read only access mode set to r.
  func (o TxOptions) WithReadOnly(r bool) TxOptions

  // WithDeferrable returns a shallow copy of the pool
  // with the transaction deferrable mode set to d.
  func (o TxOptions) WithDeferrable(d bool) TxOptions
```
- adds type RetryOptions struct
```go
  // WithDefault sets the rule for all conditions to rule.
  func (o RetryOptions) WithDefault(rule RetryRule) RetryOptions

  // WithCondition sets the retry rule for the specified condition.
  func (o RetryOptions) WithCondition(RetryCondition, RetryRule) RetryOptions
```
- adds type RetryRule struct
```go
  // WithAttempts sets the rule's attempts. attempts must be greater than zero.
  func (r RetryRule) WithAttempts(attempts int) RetryRule

  // WithBackoff returns a copy of the RetryRule with backoff set to fn.
  func (r RetryRule) WithBackoff(fn RetryBackoff) RetryRule
```
- interfaces Tryer, Executor, and Tx are removed and replaced with
  concrete types.

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -42,25 +42,10 @@ func max(a, b int) int {
 	return b
 }
 
-// Pool is a connection pool.
-type Pool interface {
-	Executor
-	Trier
-
-	// Acquire returns a connection from the pool
-	// blocking until a connection is available.
-	// Acquired connections must be released to the pool when no longer needed.
-	Acquire(context.Context) (PoolConn, error)
-
-	// Close closes all connections in the pool.
-	// Calling close blocks until all acquired connections have been released,
-	// and returns an error if called more than once.
-	Close() error
-}
-
-type pool struct {
+// Pool is a connection pool and is safe for concurrent use.
+type Pool struct {
 	isClosed bool
-	mu       sync.RWMutex // locks isClosed
+	mu       *sync.RWMutex // locks isClosed
 
 	// A buffered channel of connections ready for use.
 	freeConns chan *reconnectingConn
@@ -71,6 +56,9 @@ type pool struct {
 	maxConns int
 	minConns int
 
+	txOpts    TxOptions
+	retryOpts RetryOptions
+
 	cfg *connConfig
 
 	typeIDCache   *cache.Cache
@@ -79,7 +67,7 @@ type pool struct {
 }
 
 // Connect a pool of connections to a server.
-func Connect(ctx context.Context, opts Options) (Pool, error) { // nolint:gocritic,lll
+func Connect(ctx context.Context, opts Options) (*Pool, error) { // nolint:gocritic,lll
 	return ConnectDSN(ctx, "", opts)
 }
 
@@ -92,7 +80,7 @@ func Connect(ctx context.Context, opts Options) (Pool, error) { // nolint:gocrit
 //     edgedb://user:password@host:port/database?option=value.
 //
 // The following options are recognized: host, port, user, database, password.
-func ConnectDSN(ctx context.Context, dsn string, opts Options) (Pool, error) { // nolint:gocritic,lll
+func ConnectDSN(ctx context.Context, dsn string, opts Options) (*Pool, error) { // nolint:gocritic,lll
 	minConns := defaultMinConns
 	if opts.MinConns > 0 {
 		minConns = int(opts.MinConns)
@@ -115,10 +103,16 @@ func ConnectDSN(ctx context.Context, dsn string, opts Options) (Pool, error) { /
 		return nil, err
 	}
 
-	p := &pool{
+	p := &Pool{
+		mu:       &sync.RWMutex{},
 		maxConns: maxConns,
 		minConns: minConns,
 		cfg:      cfg,
+		txOpts: TxOptions{
+			isolation:  RepeatableRead,
+			readOnly:   false,
+			deferrable: false,
+		},
 
 		freeConns:      make(chan *reconnectingConn, minConns),
 		potentialConns: make(chan struct{}, maxConns),
@@ -158,7 +152,7 @@ func ConnectDSN(ctx context.Context, dsn string, opts Options) (Pool, error) { /
 	return p, nil
 }
 
-func (p *pool) newConn(ctx context.Context) (*reconnectingConn, error) {
+func (p *Pool) newConn(ctx context.Context) (*reconnectingConn, error) {
 	conn := &reconnectingConn{
 		conn: &baseConn{
 			cfg:           p.cfg,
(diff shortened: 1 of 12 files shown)


fmoor:
This change is based on #119 and is WIP until the upstream branch is merged.

1st1:
@tailhook Paul, please take a look.

tailhook:
I haven't look at the code yet, but here is my plan for the Rust API:

```
fn with_retry_options(opt: RetryOptions) {}
fn with_transation_options(opt: TransactionOptions) {}
```
So generally if you want to update single setting it's a bit more verbose:
```rust
# in python: con1 = con.with_transaction_options(deferred=True)
let con1 = con.with_transaction_options(
  con.get_transaction_options().with_deferred(true))
# or maybe this way, but probably not, as this is a backwards compatibility nightmare:
let con1 = con.with_transaction_options(TransactionOptions {
  deferred: true,
  ..con.get_transaction_options()
})
```
But in practice I think in statically typed languages it will be more like this:
```
const default_options: TransactionOptions = TransactionOptions::new();
const myopt1: TransactionOptions = default_options.with_deferred(true);

fn do_something(con: Connection) {
  let my_con = con.with_transaction_options(myopt1);
}
```

@1st1 was kinda against that idea in Python, but in statically typed languages this is a bit of a nightmare (even without considering forward compatibility):
```
const isolation = ..;
const deferred = false;
const readonly = false;

let con1 = con
  .withTxIsolation(isolation)
  .withDeferred(deferred)
  .withReadonly(readonly)
```

And other options will be added in the future, like `withReplicaOptions` (and adding more options is a nightmare for backwards compatibility for potential middleware libraries for EdgeDB).

pool.go
```diff
@@ -42,25 +42,10 @@ func max(a, b int) int {
...
-	// Close closes all connections in the pool.
-	// Calling close blocks until all acquired connections have been released,
-	// and returns an error if called more than once.
-	Close() error
-}
-
-type pool struct {
+// Pool is a connection pool and is safe for concurrent use.
+type Pool struct {
 	isClosed bool
```
tailhook, line 47: This should be shared I think

pool.go
```diff
@@ -71,6 +56,9 @@ type pool struct {
 	maxConns int
 	minConns int
```
tailhook, line 57: I feel a little bit uneasy because these aren't shared. But they can't be modified, right?

fmoor: By shared do you mean publicly accessible? The user can not modify the pool's min/max connection values after connecting the pool. The pool options implementation is not final. That will be revisited when we have the connection options strategy worked out.

pool.go
fmoor: You're saying there should be a `pool.IsClosed()` method?

pool.go
tailhook: No, by "shared" I mean shared between multiple copies created by `withSomething` method. If they can't be modified, its fine to copy.

pool.go
fmoor: It should be shared between copies of pools! Got it. Yes, good catch!

fmoor: changed

tailhook (approved):

1st1 (approved):
This looks good to me. But before merging, please wait for @elprans to also take a quick look at the new public API.

elprans (commented):
The API looks good to me in general, except that I don't think that we need to distinguish between `SeralizationError` and `Deadlock`, because those are basically the same class of an error.

tailhook:
> The API looks good to me in general, except that I don't think that we need to distinguish between `SeralizationError` and `Deadlock`, because those are basically the same class of an error.

My speculation is that difference is huge: deadlock means there is something which is locked, conflicting lock could take arbitrary amount of time (e.g. if this is a huge batch update), so if you see deadlocks in the log, you might bump interval on deadlock specifically. On the other hand, "serialization error" means something already happened, so even repeating immediately will work, so interval here is just to avoid DoS if there are too many concurrent transactions. Although, I don't have experience in tweaking these, so it may not be so important in practice.

elprans:
> My speculation is that difference is huge: deadlock means there is something which is locked

No, this is not what a `DeadlockError` in Postgres means.  *All* mutations take locks, and so if you have concurrent transactions that update the same rows they'd block.  The difference between a `SerializationError` and a `DeadlockError` is in the ordering of update operations.

DeadlockError:

```
-------------            -------------
Transaction 1            Transaction 2
-------------            -------------

START TRANSACTION;       START TRANSACTION;
# takes lock on A
UPDATE A;
                         # takes lock on B
                         UPDATE B;
# waits for TX 2
UPDATE B;
                         # waits for TX 1
                         UPDATE A;
```

Both transactions above block on each other and Postgres raises `DeadlockError` *immediately* when it sees `UPDATE A` in Transaction 2.


SerializationError:

```
-------------            -------------
Transaction 1            Transaction 2
-------------            -------------

START TRANSACTION;       START TRANSACTION;
# takes lock on A
UPDATE A;
                         # waits for TX 1
                         UPDATE A;
COMMIT;
                         # SerializationError
                         # because TX 1 updated A
```


No error:


```
-------------            -------------
Transaction 1            Transaction 2
-------------            -------------

START TRANSACTION;       START TRANSACTION;
# takes lock on A
UPDATE A;
                         # waits for TX 1
                         UPDATE A;
ROLLBACK;
                         # no error, because TX 1
                         # rolled back
                         COMMIT;
```


As you can see, both `DeadlockError` and `SerializationError` arise from write concurrency and are quite similar.  I don't think we need to distinguish between the two.

fmoor:
If we are merging the deadlock and serialization conditions should it be called `TransactionError`?

elprans:
`TransactionError` seems too general, as there could be other non-retryable transaction errors.  Perhaps `TransactionConcurrencyError` or `TransactionConflictError`?

tailhook:
We also should probably make a common base for Deadlock and Serialization error classes if we think that this distinction is useless.

elprans:
> We also should probably make a common base

Yes, this is what Frederick and I were referring to above.  `TransactionConflictError` seems like a good name to me.

Issue #126, opened 2021-04-07
connecting with empty options doesn't work


fmoor:
needs debugging

```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/edgedb/edgedb-go"
)

func main() {
	ctx := context.Background()
	con, err := edgedb.ConnectOne(ctx, edgedb.Options{})
	if err != nil {
		log.Fatal(err)
	}

	var result []struct {
		Name string `edgedb:"name"`
	}

	err = con.Query(ctx, "SELECT sys::Database{name}", &result)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(result)
}
```

```bash
$ go run main.go
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/edgedb/edgedb-go.(*baseConn).fallThrough(0xc0000d2480, 0xc000090100, 0xc000096101, 0x0)
        /home/fmoor/src/edgedb-go/fallthrough.go:39 +0x5b7
github.com/edgedb/edgedb-go.(*baseConn).connect(0xc0000d2480, 0xc000090100, 0xc000092180, 0x0, 0x0)
        /home/fmoor/src/edgedb-go/connect.go:99 +0x54b
github.com/edgedb/edgedb-go.connectWithTimeout(0x756eb0, 0xc0000ac000, 0xc0000d2480, 0xc0000cc100, 0x0, 0x0)
        /home/fmoor/src/edgedb-go/edgedb.go:92 +0x88f
github.com/edgedb/edgedb-go.(*reconnectingConn).reconnect(0xc0000cc140, 0x756eb0, 0xc0000ac000, 0xc0000a6840, 0x0)
        /home/fmoor/src/edgedb-go/reconnect.go:97 +0x17e
github.com/edgedb/edgedb-go.ConnectOneDSN(0x756eb0, 0xc0000ac000, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/fmoor/src/edgedb-go/conn.go:90 +0x2c5
github.com/edgedb/edgedb-go.ConnectOne(...)
        /home/fmoor/src/edgedb-go/conn.go:60
main.main()
        /home/fmoor/src/edgedb-go/cmd/sample/main.go:13 +0xba
exit status 2
```

PR #127, opened 2021-04-08
remove unused connection attribute

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/126 which didn't have to do with options.

When running edgedb/edgedb-go against a development server (`edb server`) the server sent a [ParameterStatus message](https://www.edgedb.com/docs/internals/protocol/messages#parameterstatus) that the packaged server builds don't seem to send. The bit of code that is changed in this PR is not usually exercised because the tests don't have a way of triggering server initiated messages.

### DIFF
--- a/edgedb.go
+++ b/edgedb.go
@@ -92,8 +92,6 @@ type baseConn struct {
 	inCodecCache  *cache.Cache
 	outCodecCache *cache.Cache
 
-	serverSettings map[string]string
-
 	cfg *connConfig
 }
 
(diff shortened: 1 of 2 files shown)


fmoor:
Sorry @1st1 requesting a review was a reflex. Its probably not needed for this.

1st1:
> Sorry @1st1 requesting a review was a reflex. Its probably not needed for this.

NP ;)

Issue #124, opened 2021-03-26
make id field opt-in


fmoor:
Currently the id field is always included in query results which requires users to declare an id field in their result structs even if they aren't going to use it. The id field should be opt-in instead of required.j

PR #128, opened 2021-04-08
set EXPLICIT_OBJECTIDS header

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/124

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -25,16 +25,51 @@ import (
 	"github.com/xdg/scram"
 )
 
-const (
-	protocolVersionMajor uint16 = 0
-	protocolVersionMinor uint16 = 10
+var (
+	protocolVersionMin = version{0, 9}
+	protocolVersionMax = version{0, 10}
 )
+
+type version struct {
+	major uint16
+	minor uint16
+}
+
+func (v version) gt(other version) bool {
+	switch {
+	case v.major > other.major:
+		return true
+	case v.major < other.minor:
+		return false
+	default:
+		return v.minor > other.minor
+	}
+}
+
+func (v version) gte(other version) bool {
+	if v == other {
+		return true
+	}
+
+	return v.gt(other)
+}
+
+func (v version) lt(other version) bool {
+	switch {
+	case v.major < other.major:
+		return true
+	case v.major > other.minor:
+		return false
+	default:
+		return v.minor < other.minor
+	}
+}
 
 func (c *baseConn) connect(r *buff.Reader, cfg *connConfig) error {
 	w := buff.NewWriter(c.writeMemory[:0])
 	w.BeginMessage(message.ClientHandshake)
-	w.PushUint16(protocolVersionMajor)
-	w.PushUint16(protocolVersionMinor)
+	w.PushUint16(protocolVersionMax.major)
+	w.PushUint16(protocolVersionMax.minor)
 	w.PushUint16(2) // number of parameters
 	w.PushString("database")
 	w.PushString(cfg.database)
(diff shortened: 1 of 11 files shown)


fmoor:
WIP until there is a nightly build that includes https://github.com/edgedb/edgedb/pull/2443

connect.go
```diff
@@ -25,16 +25,51 @@ import (
...
+		return true
+	case v.major < other.minor:
+		return false
+	default:
+		return v.minor > other.minor
+	}
+}
+
+func (v version) gte(other version) bool {
+	if v == other {
```
1st1, line 50: Does `==` compares all fields or just the pointers?

internal/codecs/nil.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/edgedb/edgedb-go/internal/buff"
+	"github.com/edgedb/edgedb-go/internal/descriptor"
+	types "github.com/edgedb/edgedb-go/internal/edgedbtypes"
+)
+
+type nilDecoder struct{}
+
+func (c nilDecoder) DescriptorID() types.UUID { return descriptor.NilID }
+
+func (c nilDecoder) Decode(r *buff.Reader, out unsafe.Pointer) {}
```
1st1, line 31: Hm, this seems to be something that doesn't decode literal `NULL`/empty sets into `nil`, it's just a codec that does nothing, is that right?

granularflow.go
```diff
@@ -110,9 +111,15 @@ func (c *baseConn) pesimistic(r *buff.Reader, q *gfQuery) error {
 }
 
 func (c *baseConn) prepare(r *buff.Reader, q *gfQuery) (idPair, error) {
+	headers := copyHeaders(q.headers)
+
+	if c.protocolVersion.gte(version{0, 10}) {
```
1st1, line 116: I think we should set some internal bool flag like `c.ver_0_10` to avoid this function call.

granularflow.go
```diff
@@ -190,9 +197,15 @@ func (c *baseConn) describe(r *buff.Reader, q *gfQuery) (descPair, error) {
 			// input descriptor
 			descs.in = descriptor.Pop(r.PopSlice(r.PopUint32()))
 
-			// output descriptor
-			r.Discard(16) // descriptor ID
-			descs.out = descriptor.Pop(r.PopSlice(r.PopUint32()))
+			// output descriptor ID
+			outID := r.PopUUID()
+
+			if outID == descriptor.NilID {
```
1st1, line 203: Hm, I'm not sure I understand how this check works... outID comes from our protocol, right?

connect.go
fmoor: If `v` and `other` were pointers then `==` would only compare the pointers, but they are both passed by value to this function so the struct fields are compared.

internal/codecs/nil.go
fmoor: Correct it decodes empty blocks corresponding to descriptor id 0. I'm open to other names. I didn't think too hard about that one.

granularflow.go
fmoor: correct. from the [docs](https://www.edgedb.com/docs/internals/protocol/typedesc#type-descriptors):

> There is one special type with type id of zero: 00000000-0000-0000-0000-000000000000. The describe result of this type contains zero blocks. It’s used when a statement returns no meaningful results, e.g. the CREATE DATABASE example statement.

fmoor: maybe I should call it `descriptor.ZeroID` :thinking: then `nilDecoder` would become `zeroDecoder`... or something.

internal/codecs/nil.go
1st1: I'd call it "NoOpDecoder"

granularflow.go
1st1: Ah, I get it now, thanks

1st1 (approved):
Approving this so you can commit when the nits are addressed

internal/codecs/nil.go
fmoor: perfect! I came to the same conclusion independently

changed

PR #136, opened 2021-06-23
differentiate missing from zero values

### LINKED ISSUE #135: require nullable values for missing fields
Currently unset properties are skipped when decoding to structs in go leaving whatever value was there before the call to `Query()`. The driver should require the out type to implement some sort of nullable interface or return an error when this happens.

fmoor:
Differentiate between a missing value and a zero value in query results. This change adds:

- Optional scalar types
- Arrays don't have an optional type instead nil is used to represents a missing array and an empty slice represents an empty array.
- An Optional struct that can be embeded in user defined structs to make them optional.
- an OptionalUnmarshaler interface for custom unmarshalers.

fixes #135

### DIFF
--- a/internal/codecs/object.go
+++ b/internal/codecs/object.go
@@ -24,8 +24,14 @@ import (
 	"github.com/edgedb/edgedb-go/internal/buff"
 	"github.com/edgedb/edgedb-go/internal/descriptor"
 	types "github.com/edgedb/edgedb-go/internal/edgedbtypes"
-	"github.com/edgedb/edgedb-go/internal/marshal"
+	"github.com/edgedb/edgedb-go/internal/introspect"
 )
+
+// OptionalDecoder is used when decoding optional shape fields.
+type OptionalDecoder interface {
+	DecodeMissing(unsafe.Pointer)
+	DecodePresent(unsafe.Pointer)
+}
 
 func buildObjectDecoder(
 	desc descriptor.Descriptor,
@@ -41,7 +47,7 @@ func buildObjectDecoder(
 	fields := make([]*DecoderField, len(desc.Fields))
 
 	for i, field := range desc.Fields {
-		sf, ok := marshal.StructField(typ, field.Name)
+		sf, ok := introspect.StructField(typ, field.Name)
 		if !ok {
 			return nil, fmt.Errorf(
 				"expected %v to have a field named %q", path, field.Name,
@@ -53,9 +59,17 @@ func buildObjectDecoder(
 			sf.Type,
 			path.AddField(field.Name),
 		)
-
 		if err != nil {
 			return nil, err
+		}
+
+		if !field.Required {
+			if _, isOptional := child.(OptionalDecoder); !isOptional {
+				return nil, fmt.Errorf(
+					"expected %v at %v.%v to implement OptionalUnmarshaler "+
+						"because the field is not required",
+					sf.Type, path, field.Name)
+			}
 		}
 
 		fields[i] = &DecoderField{
@@ -65,7 +79,13 @@ func buildObjectDecoder(
 		}
 	}
 
-	return &objectDecoder{desc.ID, fields}, nil
+	decoder := objectDecoder{desc.ID, fields}
+
+	if reflect.PtrTo(typ).Implements(optionalUnmarshalerType) {
+		return &optionalObjectDecoder{decoder, typ}, nil
+	}
+
+	return &decoder, nil
 }
 
 type objectDecoder struct {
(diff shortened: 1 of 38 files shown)


tutorial_test.go
```diff
@@ -32,12 +33,30 @@ type Person struct {
 	LastName  string `edgedb:"last_name"`
 }
 
+type OptionalYear struct {
+	year    int64
+	missing bool
+}
```
1st1, line 39: I don't like this API :(

Perhaps we should take a similar route to https://github.com/markphelps/optional -- define an "optional" struct for all primitive types that EdgeDB can return with a `Get()` method defined on those structs.

So instead of defining your own unmarshaller and struct every time you see an optional scalar you could use a predefined struct, e.g. `edgedb.OptionalInt64`.

1st1 (changes requested):
The machinery of optionals decoding looks good to me. But i think we need to add a higher-level set of types to ease the pain of dealing with optionals. They can use the new UnmarshalMissing/UnmarshalPresent mechanism under the hood.

tutorial_test.go
fmoor: We can do that in addition to what is here as another level of abstraction, but I think we need to expose something like this also because 1) shapes can be optional and 2) we support user defined scalar codecs.

1st1: Yeah, I'm not suggesting that we should hide the API, just make it more convenient to work with.

1st1: Question: can we declare our own `Optional` struct and let people embed it into optional structs? Like this (in psudo-code):

```go
// edgedb-go:

type Optional struct {
    __loaded__ bool
}

func (l *Optional) isEmpty() bool {
   return l.__loaded__
}
```

so that users would be able to do this:

```
type Movie struct {
   edgedb.Optional
   ....
}

// and later do

if movie.isEmpty() { ... }
```

?

Here's a runnable example: https://play.golang.org/p/mXK2jNDzhLQ

fmoor: Yes we can. Good idea!

1st1: Then with this and with the scalar optional wrappers dealing with optionals should become easier

fmoor:
Updated to follow cardinality enum changes in https://github.com/edgedb/edgedb/pull/2635. The tests won't pass until the server changes make it into nightly.

tutorial_test.go
tailhook: Hm, I think common pattern in go is to use pointers? I.e. if you need optional int, you use `field *int`. And then you check whether `x.field != nil` for the presence of the value.

fmoor: That is a pattern, but I'm not sure it is THE pattern. The stdlib uses structs like this for the same purpose in [database/sql](https://golang.org/pkg/database/sql/#NullBool). When using pointers it is easy to forget to check if the pointer is `nul`. This API forces users to think more about what they are doing and hopefully makes it more obvious to readers what is going on.

elprans: The problem with scalar pointers in structs is that you cannot take an address of an expression, so working with structs becomes a nightmare, e.g.:

```go
bool_field := some_bool_expr
struct = Struct{bool_field: &bool_field}
```

instead of

```
Struct{bool_field: some_bool_expr}
```

This obviously becomes a lot less pleasant as the number of fields grows.  Another pitfall is copying, as a copy of a Struct will reuse pointers to values instead of copying the values, so you need to write explicit copy functions.

tailhook: @elprans 
1. This is mostly for rows returned from the database, right? So this ergonomic doesn't play role. For input: you rarely use optional variables in queries, and if you do, they go from somewhere, not constant right right in `.query()` call (you would put the constant in the query itself, probably, or just don't do it optional, since you know the value).
2. With copying I think there is the same deep vs shallow copy for arrays, right? So generally this is a known quirk in go. On the other hand, I don't think "fetch then clone then edit structure" is a common pattern. Even just editing a structure in place is not common, unless we encourage using EdgeDB in ORM style. And even when using ORM cloning is somewhat rarely used in my practice (part of that is confusing identity, which has to be rewritten during the clone anyway, so special clone function anyway).

Note that not forgetting to fill `__loaded__` is generally a more real problem if creating such structures would be a common code. And not forgetting to check `__loaded__` can be a bigger problem (i.e. a security issue) than occasionally crashing on `nil` (there should be a go lint to check the latter, no?)

1st1: Imo we don't need the pointers based api and my suggestion about adding optional versions of the primitive types + and optional struct marker should suffice.

fmoor:
@1st1 I think this is ready for review again.

types_test.go
```diff
@@ -2337,3 +3246,179 @@ func TestReceiveSet(t *testing.T) {
...
+		type WrongType struct {
+			SimpleScalar int64 `edgedb:"simple_scalar"`
+		}
+
+		var result2 WrongType
+		e = tx.QueryOne(ctx, `
+			SELECT Sample { simple_scalar } LIMIT 1`,
+			&result2,
+		)
+		assert.EqualError(t, e, "edgedb.UnsupportedFeatureError: "+
```
elprans, line 3415: This should be `InvalidArgumentError`.  `UnsupportedFeatureError` is a server error.  Binding-originated errors should use the `ClientError` descendants.

fmoor: changed

internal/codecs/object.go
```diff
@@ -53,19 +59,33 @@ func buildObjectDecoder(
...
 		)
-
 		if err != nil {
 			return nil, err
 		}
 
+		if !field.Required {
+			if _, isOptional := child.(OptionalDecoder); !isOptional {
+				return nil, fmt.Errorf(
+					"expected %v at %v.%v to implement OptionalUnmarshaler "+
```
elprans, line 69:
```suggestion
					"expected %v at %v.%v to implement OptionalDecoder "+
```

I'd also replace (un)marshaller with (de)coder everywhere to be consistent.

fmoor: I thought about using Decoder. I went with Unmarshaller instead because it is what the stdlib uses https://pkg.go.dev/encoding

elprans: Why is the interface named `OptionalDecoder` then?

fmoor: Probably because `Decoder` is used in `internal/` types and I mixed them up when I wrote this. I think the interface should be `OptionalUnmarshaller`.

fmoor: It looks like I actually have both an `OptionalUnmarshaler` and `OptionalDecoder` interface and they have different methods. Let me thing about this and push a change :thinking:

fmoor: On looking at this again, `type OptionalDecoder interface` is an internal API that is implemented by Decoder types defined in `internal/`. The suggested change is correct. I have updated the PR.

> I'd also replace (un)marshaller with (de)coder everywhere to be consistent.

I think that (Un)Marshaler is a more helpful name for the public API because of its use in the stdlib.

fmoor: The type cast to `OptionalDecoder` is checking if the internal decoder  knows how to treat the user's out value as optional. If not the error message reports to the user that their out type needs to implement `OptionalUnmarshaler`. This error message is correct.

fmoor: Here is the assertion that tests this behavior https://github.com/edgedb/edgedb-go/pull/136/files#diff-02776af2e303aa4b13cf883561bfd1aa91a940389c39aae1b7abbaecc0d68092R3416

1st1 (approved):

PR #137, opened 2021-06-30
add TLS support

fmoor:


fmoor:
If we want to use `OptionalBool` for the `edgedb.Options.VerifyHostname` argument to `Connect()` then this PR is blocking on https://github.com/edgedb/edgedb-go/pull/136

elprans:
Let's port the definition of optionals from #136 to unblock and avoid rushing #136.

fantix (approved):
I'm not familiar with Golang, but the logic looks correct to me. 👍

connect.go
```diff
@@ -152,6 +153,14 @@ func (c *baseConn) connect(r *buff.Reader, cfg *connConfig) error {
 		}
 	}
 
+	_, isTLS := c.conn.(*tls.Conn)
+	if !isTLS && c.protocolVersion.gte(version{0, 11}) {
+		_ = c.close()
+		return &clientConnectionError{msg: fmt.Sprintf(
+			"server clamming to use protocol version %v.%v without using TLS",
```
elprans, line 160:
```suggestion
			"server claims protocol version %v.%v without using TLS",
```

windows.go
```diff
@@ -18,4 +18,12 @@
...
 package edgedb
 
+import "crypto/x509"
+
 var defaultHosts = []string{"localhost"}
+
+func getCertPool() (*x509.CertPool, error) {
+	// x509.SystemCertPool() doesn't work on Windows.
+	// https://github.com/golang/go/issues/16736
+	return x509.NewCertPool(), nil
```
elprans, line 28: I think we should bundle the Mozilla CA bundle for parity with other bindings. certifi has a [copy](https://github.com/certifi/python-certifi/blob/master/certifi/cacert.pem)

connutils.go
```diff
@@ -448,6 +490,54 @@ func parseConnectDSNAndArgs(
 		waitUntilAvailable = 30 * time.Second
 	}
 
+	roots, err := getCertPool()
```
fmoor, line 493: We should probably not include system certs when the user provides their own certificate right?

elprans: I think so.

fmoor: changed

windows.go
fmoor: changed

connect.go
elprans: This doesn't seem to have been changed in the latest diff.

elprans (approved):
LGTM!

PR #138, opened 2021-07-10
Update doc_test.go

divan:
Fixing two issues with this example:

1. `inserted` should be a pointer (current example yields error `the "out" argument must be a pointer, got struct { id edgedbtypes.UUID }`
2. positional arguments seems to start with 0 (current example yields `edgedb.QueryError: missing $0 positional argument` error)


fmoor:
Good catch @divan! Thanks for the PR. Once you sign the License Agreement I'll merge this.

PR #139, opened 2021-07-12
Reduce memory used by test suite

fmoor:
The test suite frequently runs out of memory on Windows in github actions.


Makefile
```diff
@@ -4,7 +4,7 @@ lint:
 	golangci-lint run
 
 test:
-	go test -race -bench=$$^ -timeout=5m ./...
+	go test -v -race -bench=$$^ -timeout=5m ./...
```
elprans, line 7: Running with `-v` reduces memory usage?

fmoor: no this is WIP. I was checking to see if there is a particular test that always causes the crash. -v prints the test names as they run.

elprans (approved):

PR #146, opened 2021-08-02
Add QuerySingle() and deprecate QueryOne()

fmoor:
QuerySingJSON() is also added and QueryOneJSON() is also deprecated.

### DIFF
--- a/internal/cardinality/cardinality.go
+++ b/internal/cardinality/cardinality.go
@@ -18,14 +18,14 @@ package cardinality
 
 // Cardinalities
 const (
-	NoResult = 0x6e
-	Single   = 0x6f
-	Many     = 0x6d
+	NoResult  = 0x6e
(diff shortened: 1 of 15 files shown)


internal/cardinality/cardinality.go
```diff
@@ -19,13 +19,13 @@ package cardinality
 // Cardinalities
 const (
 	NoResult = 0x6e
-	One      = 0x6f
+	Single   = 0x6f
```
elprans, line 22: This should be called `AtMostOne` to be consistent with the server enum.  `single` and `multi` are simplified cardinality indicators for the full range of possible cardinalities, where `single` is compatible with {`zero`, `at_most_one`, `one`}, and `multi` is compatible with all of them.

fmoor: changed

elprans (approved):

PR #157, opened 2021-09-07
Make RetryingTX() retry on START TX failure

fmoor:


fantix (approved):
From the code it makes sense to me but please pardon my poor Go knowledge and weight @elprans's review.

elprans (approved):

Issue #159, opened 2021-09-09
RetryingTx() should retry on all network errors


fmoor:
Currently only edgedb errors with the `SHOULD_RETRY` tag are retried.

PR #160, opened 2021-09-17
RectryingTx reconnects on all network errors

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/159

A new abstraction is introduced that wraps the net.Conn. This socket abstraction 1) closes itself when any read or write fails and 2) wraps all errors returned in an appropriate EdgeDB error type.

The baseConn type is renamed to protocolConn.


soc.go
```diff
@@ -0,0 +1,150 @@
+// This source file is part of the EdgeDB open source project.
```
elprans, line 1: I'd name this sock.go

internal/buff/write_test.go
```diff
@@ -1,265 +0,0 @@
-// This source file is part of the EdgeDB open source project.
```
elprans, line 1: Is this removed on purpose?

fmoor: Yes. The testing strategy going forward is to only maintain functional tests. This change broke the tests in this file so I got rid of it instead of fixing them.

soc.go
fmoor: changed

elprans (approved):

PR #161, opened 2021-09-21
Retry readonly queries outside of transactions

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/158

Other minor changes:
- fixed bug in type ID caching. The key used to retrieve entries from the cache had different pieces of information from the key used to cache entries.
- fixed QueryOneJSON which was sending queries as format.Binary

### DIFF
--- a/granularflow.go
+++ b/granularflow.go
@@ -35,7 +35,7 @@ func (c *protocolConnection) execGranularFlow(
 	r *buff.Reader,
 	q *gfQuery,
 ) error {
-	ids, ok := c.getTypeIDs(q)
+	ids, ok := c.getCachedTypeIDs(q)
 	if !ok {
 		return c.pesimistic(r, q)
 	}
@@ -192,7 +192,7 @@ func (c *protocolConnection) prepare(r *buff.Reader, q *gfQuery) error {
 	for r.Next(done.Chan) {
 		switch r.MsgType {
 		case message.PrepareComplete:
-			ignoreHeaders(r)
+			c.cacheCapabilities(q, decodeHeaders(r))
 			r.Discard(1) // cardianlity
 			ids = idPair{in: [16]byte(r.PopUUID()), out: [16]byte(r.PopUUID())}
 		case message.ReadyForCommand:
@@ -212,7 +212,7 @@ func (c *protocolConnection) prepare(r *buff.Reader, q *gfQuery) error {
 		return r.Err
 	}
 
-	c.putTypeIDs(q, ids)
+	c.cacheTypeIDs(q, ids)
 
 	return err
 }
@@ -244,7 +244,7 @@ func (c *protocolConnection) describe(
 	for r.Next(done.Chan) {
 		switch r.MsgType {
 		case message.CommandDataDescription:
-			descs, err = c.decodeCommandDataDescriptionMsg(r, q)
+			descs, _, err = c.decodeCommandDataDescriptionMsg(r, q)
 		case message.ReadyForCommand:
 			decodeReadyForCommandMsg(r)
 			done.Signal()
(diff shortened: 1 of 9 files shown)


fmoor:
This change is based on https://github.com/edgedb/edgedb-go/pull/160. WIP until the upstream change is merged.

granularflow.go
```diff
@@ -374,7 +387,9 @@ func (c *baseConn) optimistic(
 		case message.CommandComplete:
 			decodeCommandCompleteMsg(r)
 		case message.CommandDataDescription:
-			descs, err = decodeCommandDataDescriptionMsg(r, q, c)
+			var headers msgHeaders
+			descs, headers, err = c.decodeCommandDataDescriptionMsg(r, q)
+			c.cacheHeaders(q, headers)
```
fmoor, line 392: It is unclear to me from the [protocol documentation](https://www.edgedb.com/docs/internals/protocol/overview#granular-flow) if this is correct. The documentation says that if optimistic execute is run with erroneous IDs that the server will prepare the query and return a CommandDataDescription message. However, the CommandDataDescription message has no known headers listed. Will the headers generated from preparing the query be returned with the CommandDataDescription message in this case? The headers from PrepareComplete are important because they indicate weather a query is read only or not.

elprans: It's a bug.  `CommandDataDescription` should be equivalent to `PrepareComplete`.  I'll make a fix.

elprans:
Please rebase.

granularflow.go
fmoor: > `CommandDataDescription` should be equivalent to `PrepareComplete`

As far as the headers they may return?

granularflow.go
```diff
@@ -192,7 +192,7 @@ func (c *protocolConnection) prepare(r *buff.Reader, q *gfQuery) error {
 	for r.Next(done.Chan) {
 		switch r.MsgType {
 		case message.PrepareComplete:
-			ignoreHeaders(r)
+			c.cacheHeaders(q, decodeHeaders(r))
```
elprans, line 195: I would be granular here and cache the returned capabilities word explicitly.  Caching all headers indiscriminately is likely incorrect as you might accidentally rely on a piece of data that isn't actually stable.

transactable.go
```diff
@@ -108,6 +108,49 @@ func (c *transactableConn) QuerySingleJSON(
...
+		}
+
+		err = c.reconnectingConn.granularFlow(ctx, q)
+
+	Error:
+		// q is a read only query if the server sent zero headers when q was
+		// prepared. Read only queries are retryable.
+		headers, _ := c.getCachedHeaders(q)
+
+		if len(headers) == 0 &&
```
elprans, line 135: This is incorrect.  The query "readonliness" is indicated by a specific header: https://github.com/edgedb/edgedb/blob/02f95c12b35f51322050e7f3043f2cd159ef2a97/edb/server/protocol/binary.pyx#L115

elprans: The actual capability flags are defined here: https://github.com/edgedb/edgedb/blob/02f95c12b35f51322050e7f3043f2cd159ef2a97/edb/server/compiler/enums.py#L40-L46

Essentially, the absence of the CAPABILITIY header or its value of `0` is an indicator that the query has no side effects.

elprans: Actually, the fixed server would always return the `CAPABILITY` header, so if it's absent, it's the old server which forgets to send the header and so we should assume it isn't safe to retry.

granularflow.go
elprans: As far as the returned `CAPABILITY` header.

granularflow.go
fmoor: changed

transactable.go
fmoor: changed

elprans (approved):

Issue #178, opened 2021-10-20
Update environement variable implementation


fmoor:
Implement relevant changes from https://github.com/edgedb/edgedb/discussions/3049

PR #179, opened 2021-10-20
Update setting lookup from environment variables

fmoor:
fix https://github.com/edgedb/edgedb-go/issues/178

### DIFF
--- a/connutils.go
+++ b/connutils.go
@@ -60,14 +60,14 @@ type cfgVal struct {
 }
 
 type configResolver struct {
-	host              cfgVal // string
-	port              cfgVal // int
-	database          cfgVal // string
-	user              cfgVal // string
-	password          cfgVal // OptionalStr
-	tlsCAData         cfgVal // []byte
-	tlsVerifyHostname cfgVal // OptionalBool
-	serverSettings    map[string]string
+	host           cfgVal // string
+	port           cfgVal // int
+	database       cfgVal // string
+	user           cfgVal // string
+	password       cfgVal // OptionalStr
+	tlsCAData      cfgVal // []byte
+	tlsSecurity    cfgVal // string
+	serverSettings map[string]string
 }
 
 func (r *configResolver) setHost(val, source string) error {
@@ -155,22 +155,17 @@ func (r *configResolver) setTLSCAFile(file, source string) error {
 	return nil
 }
 
-func (r *configResolver) setTLSVerifyHostname(val bool, source string) {
-	if r.tlsVerifyHostname.val != nil {
-		return
-	}
-	r.tlsVerifyHostname = cfgVal{val: val, source: source}
-}
-
-func (r *configResolver) setTLSVerifyHostnameStr(val, source string) error {
-	if r.tlsVerifyHostname.val != nil {
+func (r *configResolver) setTLSSecurity(val string, source string) error {
+	if r.tlsSecurity.val != nil {
 		return nil
 	}
-	v, err := parseVerifyHostname(val)
-	if err != nil {
-		return err
+
+	switch val {
+	case "insecure", "no_host_verification", "strict":
+	default:
+		return invalidTLSSecurity(val)
 	}
-	r.setTLSVerifyHostname(v, source)
+	r.tlsSecurity = cfgVal{val: val, source: source}
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


fmoor:
This PR is blocking on https://github.com/edgedb/shared-client-testcases/pull/2

credentials.go
```diff
@@ -25,13 +25,13 @@ import (
...
-	password       OptionalStr
-	certData       OptionalBytes
-	verifyHostname OptionalBool
+	host        OptionalStr
+	port        OptionalInt32
+	user        string
+	database    OptionalStr
+	password    OptionalStr
+	certData    OptionalBytes
+	tlsSecurity OptionalStr
```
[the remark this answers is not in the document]
fmoor: There is no such thing as an enum in go. We could subtype string if you thing that is worth it but currently the go client maps enums to string.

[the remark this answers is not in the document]
tailhook: I mean something like this: https://www.educative.io/edpresso/what-is-an-enum-in-golang
But yeah, it doesn't add to correctness in go, and is not used in performance-sensitive code, so strings are fine.

connutils.go
```diff
@@ -45,8 +45,59 @@ type connConfig struct {
...
+		}
+	}
+
+	tlsConfig := &tls.Config{
+		RootCAs:    roots,
+		NextProtos: []string{"edgedb-binary"},
+	}
+
+	security, err := getEnvVarSetting(
+		"EDGEDB_CLIENT_SECURITY", "default", "default", "insecure_dev_mode")
```
elprans, line 74: Add `strict`, which is the default.

elprans (approved):
LGTM, but please add the `strict` option to the tls security enum.

connutils.go
fmoor: I'm confused. This is the replacement for `EDGEDB_INSECURE_DEV_MODE`. Is the default really `strict` TLS?

fmoor: Shouldn't the default be to have no effect? Then whatever `tls_security` workouts out to be (who's default is `strict`) will be the end result.

elprans: Hm, you're right. The default is the same as `EDGEDB_CLIENT_TLS_SECURITY`, but `strict` should still be an option.

connutils.go
```diff
@@ -45,8 +45,60 @@ type connConfig struct {
...
+	}
+
+	tlsConfig := &tls.Config{
+		RootCAs:    roots,
+		NextProtos: []string{"edgedb-binary"},
+	}
+
+	security, err := getEnvVarSetting(
+		"EDGEDB_CLIENT_SECURITY", "default", "default", "insecure_dev_mode",
+		"strict")
```
elprans, line 75: `strict` should enable `tls_security=strict` (or error if it's set to a conflicting value).

connutils.go
```diff
@@ -615,11 +608,34 @@ func (r *configResolver) config(opts *Options) (*connConfig, error) {
...
+		"default", "insecure_dev_mode", "strict")
+	if err != nil {
+		return nil, err
+	}
+
+	switch security {
+	case "default":
+	case "insecure_dev_mode":
+		if tlsSecurity == "default" {
+			tlsSecurity = "insecure"
```
elprans, line 621: No-no, "default" means "default" for tlsSecurity also.

fmoor: I don't know how to interpret this. Can you write out something like a truth table for the interaction between these two values?

elprans:
```
CLIENT_SECURITY            CLIENT_TLS_SECURITY    result TLS_SECURITY
-------------------------  ---------------------  --------------------
default                    default                default
default                    insecure               insecure
default                    no_host_verification   no_host_verification
default                    strict                 strict
insecure_dev_mode          default                insecure
insecure_dev_mode          insecure               insecure
insecure_dev_mode          no_host_verification   no_host_verification
insecure_dev_mode          strict                 strict
strict                     default                strict
strict                     insecure               ERROR
strict                     no_host_verification   ERROR
strict                     strict                 strict
```

fmoor: I must be missing something. Where does the implementation differ from the table?

elprans: "default" `CLIENT_SECURITY` leads to "insecure" TLS mode instead of "default"

fmoor: The `CLIENT_SECURITY=default` case (line `618`) has an empty body which is a noop. Are you reading it as if that case falls through to the next case? switch cases don't fall through unless there is an explicit `fallthrough` statement.

elprans: Oh, yeah that was it, I was reading it as C. :+1:

PR #187, opened 2021-11-02
Enable QuerySingle(JSON) lower cardinality assertion

fmoor:
Make QuerySingle & QuerySingleJSON only return NoDataError on zero results if the out value's type is not an optional type.

fix https://github.com/edgedb/edgedb-go/issues/184

### DIFF
--- a/internal/edgedbtypes/numbers.go
+++ b/internal/edgedbtypes/numbers.go
@@ -32,6 +32,9 @@ func (o *Optional) Missing() bool { return !o.isSet }
 // means present.
 func (o *Optional) SetMissing(missing bool) { o.isSet = !missing }
 
+// Unset marks the value a missing
+func (o *Optional) Unset() { o.isSet = false }
+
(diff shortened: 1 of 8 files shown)


internal/edgedbtypes/numbers.go
```diff
@@ -32,6 +32,9 @@ func (o *Optional) Missing() bool { return !o.isSet }
 // means present.
 func (o *Optional) SetMissing(missing bool) { o.isSet = !missing }
 
+// Unset marks the value a missing
+func (o *Optional) Unset() { o.isSet = false }
```
elprans, line 36: Why do we need `Unset` in addition to `SetMissing`?

fmoor: Because the new query single behavior uses `type unseter interface { Unset() }` to identify optional result types. I was thinking about removing `SetMissing(bool)` and replacing it with `Set()`, but I'm not sure.

elprans (approved):

Issue #189, opened 2021-11-08
Add "Basic Usage" section to docs


colinhacks:
Currently there aren't code samples for transactions or other client methods, would be useful IMO. This also makes it consistent with the other client library docs, which make a distinction between "Basic Usage" and "API".

PR #190, opened 2021-11-10
Add a tx example

fmoor:
fix https://github.com/edgedb/edgedb-go/issues/189

### DIFF
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


fmoor:
@colinhacks the edgedb-go docs currently only use the [godoc](https://pkg.go.dev/golang.org/x/tools/cmd/godoc) output to generate the documentation for edgedb.com. godoc does not give you much control and it only has an Overview, Index, and the API sections. With the current system there isn't a way to add a basic usage section unless it was a sub heading in the Overview.

colinhacks:
Gotcha. Any reason you added this example to a test file? I think the way forward is to add a "Transctions" section to `doc.go` instead.

fmoor:
> Any reason you added this example to a test file?

Because godoc adds an [index of examples](https://pkg.go.dev/math#pkg-examples) created [this way](https://pkg.go.dev/github.com/fluhus/godoc-tricks#Examples) to the top of the documentation. This also makes the example [part of the test suite](https://go.dev/blog/examples) so it at least has to compile and run without panicking. Adding an example to `doc.go` doesn't have any checks like this. 

I don't know how our docs build process handles these examples or even if it finds them. Maybe it should append them to the Overview?

I can move the example to the Overview (`doc.go`) if that is better.

colinhacks (approved):

PR #195, opened 2021-12-10
Add support for non-file variants of `tls_ca_file` and `credentials_file`

elprans:
The client now supports `tls_ca` and `credentials` as direct options.  Also
spell `tls_ca_file` correctly as `tls_ca_file` in DSNs and move TLS
configuration options to a sub-struct in connection `Options` to future-proof it
for the possible additions of more options.

### DIFF
--- a/options.go
+++ b/options.go
@@ -42,11 +42,18 @@ type Options struct {
 	// their defaults.
 	Port int
 
+	// Credentials is a JSON string containing connection credentials.
+	//
+	// Credentials cannot be specified alongside the 'dsn' argument, Host,
+	// Port, or CredentialsFile.  Credentials will override all other
+	// credentials not present in the credentials string with their defaults.
+	Credentials []byte
+
 	// CredentialsFile is a path to a file containing connection credentials.
 	//
 	// CredentialsFile cannot be specified alongside the 'dsn' argument, Host,
-	// or Port. CredentialsFile will override all other credentials not
-	// present in the credentials file with their defaults.
+	// Port, or Credentials.  CredentialsFile will override all other
+	// credentials not present in the credentials file with their defaults.
 	CredentialsFile string
 
 	// User is the name of the database role used for authentication.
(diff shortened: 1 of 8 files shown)


options.go
```diff
@@ -85,16 +92,47 @@ type Options struct {
 	// Has no effect for single connections.
 	Concurrency uint
 
-	// Read the TLS certificate from this file
+	// Parameters used to configure TLS connections to EdgeDB server.
+	TLSOptions TLSOptions
```
fmoor, line 96: This doesn't seem to allow passing your own `tls.Config`. Maybe this could be an interface with one method `TLSConfig() (tls.Config, error)` or similar and then add that method to the TLSOptions type? Is there another plan for accepting user defined  `tls.Config`?

options.go
```diff
@@ -85,16 +92,47 @@ type Options struct {
...
+const (
+	// Default makes security mode inferred from other options
+	Default TLSSecurityMode = "default"
+	// Insecure results in no certificate verification whatsoever
+	Insecure TLSSecurityMode = "insecure"
+	// NoHostVerification enables certificate verification against CAs,
+	// but hostname matching is not performed.
+	NoHostVerification TLSSecurityMode = "no_host_verification"
+	// Strict enables full certificate and hostname verification.
+	Strict TLSSecurityMode = "strict"
```
fmoor, line 133: It feels strange to me to have `edgedb.Default` instead of a more specific name `edgedb.DefaultTLSSecurity` or something. We could easily end up with other types that have a default value.

options.go
elprans: `TLSOptions` can embed `tls.Config` in the future.  This PR doesn't deal with that just yet.

options.go
elprans: Yeah, I forgot Golang doesn't actually namespace enums.

elprans: Fixed.  I also uppercased the constants.

fmoor (approved):

PR #196, opened 2021-12-10
Fix propagation of TLS errors

elprans:
TLS errors are now wrapped into `clientConnectionFailedError` and so
`isTLSError` fails to pick them up leading to a nonsensical crash:

    server claims to use protocol version 0.13 without using TLS


error.go
```diff
@@ -199,6 +199,12 @@ func wrapAll(errs ...error) error {
 }
 
 func isTLSError(err error) bool {
+	var errEDB Error
+
+	if errors.As(err, &errEDB) {
+		err = errEDB.Unwrap()
```
fmoor, line 205: It seems like it might be more robust to use `errors.As()` with the x509 error types this function checks for. That way if the error chain changes again in the future this wont break.

fmoor (approved):

Issue #200, opened 2022-02-22
Concurrent map writes in fallthrough.go


divan:
I've just upgraded to 0.9.2 and started to see random panics witth the following stacktrace:

```
fatal error: concurrent map writes

goroutine 135 [running]:
runtime.throw({0x10503c557, 0x15})
	/usr/local/go/src/runtime/panic.go:1198 +0x54 fp=0x14000646e50 sp=0x14000646e20 pc=0x104b323d4
runtime.mapassign_faststr(0x105201400, 0x14000342690, {0x14000638660, 0x1a})
	/usr/local/go/src/runtime/map_faststr.go:294 +0x3d0 fp=0x14000646ec0 sp=0x14000646e50 pc=0x104b0fac0
github.com/edgedb/edgedb-go.(*protocolConnection).fallThrough(0x14000101680, 0x1400034b300)
	/Users/divan/pkg/mod/github.com/edgedb/edgedb-go@v0.9.2/fallthrough.go:39 +0x514 fp=0x14000646fd0 sp=0x14000646ec0 pc=0x104da6084
github.com/edgedb/edgedb-go.(*protocolConnection).authenticate(0x14000101680, 0x1400034b300, 0x140001c4f30)
	/Users/divan/pkg/mod/github.com/edgedb/edgedb-go@v0.9.2/connect.go:240 +0xc60 fp=0x14000647140 sp=0x14000646fd0 pc=0x104d9a810
github.com/edgedb/edgedb-go.(*protocolConnection).connect(0x14000101680, 0x1400034b300, 0x140001c4f30)
	/Users/divan/pkg/mod/github.com/edgedb/edgedb-go@v0.9.2/connect.go:106 +0xf20 fp=0x14000647290 sp=0x14000647140 pc=0x104d99a30
github.com/edgedb/edgedb-go.connectWithTimeout({0x1052cce60, 0x14000194810}, 0x140001c4f30, {0x14000342690, 0x140002fc7d0, 0x140002fc820, 0x140002fc870, 0x140002fc8c0})
	/Users/divan/pkg/mod/github.com/edgedb/edgedb-go@v0.9.2/edgedb.go:79 +0x310 fp=0x14000647330 sp=0x14000647290 pc=0x104da3a80
github.com/edgedb/edgedb-go.(*reconnectingConn).reconnect(0x140004d80a0, {0x1052cce60, 0x14000194810}, 0x0)
[...]

PR #201, opened 2022-02-23
Fix conccurent map writes

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/200


elprans (commented):
Can we maybe use [`sync.Map`](https://pkg.go.dev/sync#Map) for this to make things more robust?

fmoor:
> Can we maybe use [`sync.Map`](https://pkg.go.dev/sync#Map) for this to make things more robust?

The documentation for `sync.Map` discourages it unless you have a very specific use case.

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
>
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

connutils.go
```diff
@@ -31,13 +31,46 @@ import (
...
+func newServerSettings() *serverSettingsCache {
+	return &serverSettingsCache{settings: make(map[string][]byte)}
+}
+
+// serverSettingsCache is a concurrency safe map. A serverSettingsCache must
+// not be copied after first use. Use newServerSettings() instead of creating
+// serverSettingsCache manually.
+type serverSettingsCache struct {
+	settings map[string][]byte
+	mx       sync.Mutex
```
elprans, line 52: Any reason to not use an `RWMutex`?

elprans (approved):

PR #203, opened 2022-02-26
Add support for traversing into structs using '$inline' struct tag

das7pad:
Hi there!

Coming from mongodb/bson, I've adapted a pattern where I re-use field definitions for larger models by defining them once and composing them together using the `inline` tag. Like so:

<details>
<summary> Demo </summary>

---

This demo is informal and "Not a Submission.".

```go
# project
type CommonTreeFields struct {
	Id   primitive.ObjectID `json:"_id" bson:"_id"`
	Name string             `json:"name" bson:"name"`
}
type FileRef struct {
	CommonTreeFields `bson:"inline"`

	Created time.Time `json:"created" bson:"created"`
	Size    int64     `json:"size" bson:"size"`
}

# deletedFileRef
type DeletedAtField struct {
	DeletedAt time.Time `json:"deletedAt" bson:"deletedAt"`
}
type DeletedFileRef struct {
	project.FileRef `bson:"inline"`
	DeletedAtField  `bson:"inline"`
}
```

A `DeletedFileRef` has all the fields of a regular `FileRef` plus a timestamp. The bson package would fill in all the details from a map like this: `{ "_id": ..., "name": ..., "created": ..., "size": ..., "deletedAt": ... }`.

> From their docs: https://docs.mongodb.com/drivers/go/current/fundamentals/bson/#struct-tags
> If the field type is a struct or map field, the field will be flattened when marshalling and unflattened when unmarshalling.

---

</details>

I really like that pattern and would like to continue using it with edgedb.

Let me know what you think!

Edit: Per the discussion below, the keyword in edgedb will be `$inline` to avoid potential conflicts with other identifiers.
Is there a good place for documenting `$inline` as a special keyword?

### DIFF
--- a/internal/introspect/marshal.go
+++ b/internal/introspect/marshal.go
@@ -24,8 +24,15 @@ import (
 func fieldByTag(t reflect.Type, name string) (reflect.StructField, bool) {
 	for i := 0; i < t.NumField(); i++ {
 		field := t.Field(i)
-		if field.Tag.Get("edgedb") == name {
+		switch field.Tag.Get("edgedb") {
+		case name:
 			return field, true
+		case "$inline":
+			if f, ok := fieldByTag(field.Type, name); ok {
+				// Accumulate offsets from nested paths.
+				f.Offset += field.Offset
+				return f, true
+			}
 		}
 	}
 
--- a/internal/introspect/marshal_test.go
+++ b/internal/introspect/marshal_test.go
@@ -31,6 +31,78 @@ type SomeStruct struct {
 	Third  []byte `edgedb:"First"`
 }
 
+type InlinedSomeStruct struct {
+	SomeStruct `edgedb:"$inline"`
+	Zebra      string
+}
+
+type InnerOne struct {
+	One         string `edgedb:"one"`
+	OnePointOne string `edgedb:"one_point_one"`
+}
+
+type InnerTwo struct {
+	Two string `edgedb:"two"`
+}
+
+type InnerThree struct {
+	Three string `edgedb:"three"`
+}
+
+type InlinedMultipleStructs struct {
+	InnerOne   `edgedb:"$inline"`
+	InnerTwo   `edgedb:"$inline"`
+	InnerThree `edgedb:"$inline"`
+}
+
+type InlinedMultipleLayers struct {
+	InlinedSomeStruct      `edgedb:"$inline"`
+	InlinedMultipleStructs `edgedb:"$inline"`
+}
+
+func checkInlinedSomeStruct(t *testing.T, typ reflect.Type, offset uintptr) {
+	// Nested field
+	field, ok := StructField(typ, "First")
+	require.True(t, ok)
+	assert.Equal(t, "Third", field.Name)
+	assert.Equal(t, offset+16+8, field.Offset)
+
+	field, ok = StructField(typ, "Second")
+	require.True(t, ok)
+	assert.Equal(t, "Second", field.Name)
+	assert.Equal(t, offset+16, field.Offset)
+
+	// Top level field
+	field, ok = StructField(typ, "Zebra")
+	require.True(t, ok)
+	assert.Equal(t, "Zebra", field.Name)
+	assert.Equal(t, offset+16+8+24, field.Offset)
+}
+
+func checkInlinedMultipleStructs(
+	t *testing.T, typ reflect.Type, offset uintptr,
+) {
+	field, ok := StructField(typ, "one")
+	require.True(t, ok)
+	assert.Equal(t, "One", field.Name)
+	assert.Equal(t, offset+16*0, field.Offset)
+
+	field, ok = StructField(typ, "one_point_one")
+	require.True(t, ok)
+	assert.Equal(t, "OnePointOne", field.Name)
+	assert.Equal(t, offset+16*1, field.Offset)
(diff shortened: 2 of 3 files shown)


fmoor:
@das7pad, thanks for this! I like this feature. I'll discuss with the rest of the team. I think one adjustment that would be nice is to prefix the inline tag with `#` (the comment character in edgeql) or something similar so that it doesn't shadow a possible real property/link name.

fmoor:
> prefix the inline tag with `#` (the comment character in edgeql) or something similar

I just learned that properties can actually start with `#`. Would you prefix `inline` with `$`?

das7pad:
> > prefix the inline tag with `#` (the comment character in edgeql) or something similar
> 
> I just learned that properties can actually start with `#`. Would you prefix `inline` with `$`?

Thanks for your comments and bringing this up with the team!

`$inline` reads more like a variable name than a struct tag option to me. So I looked for other options.
The [`encoding/json`](https://github.com/golang/go/blob/986b04c0f12efa1c57293f147a9e734ec71f0363/src/encoding/json/encode.go#L64-L104) package handles the sole use of their special keywords (`-`/`omitempty`/`string`) by specifying an empty name followed by a comma and the special keyword `,omitempty`.
Adapted for this PR, this would be `,inline` -- which does not read too bad I think. It would fit into the mental model of using other struct tags, which is a bonus over using another special character as prefix.

I read the [docs on identifiers in EdgeQL](https://github.com/golang/go/blob/986b04c0f12efa1c57293f147a9e734ec71f0363/src/encoding/json/encode.go#L64-L104), which suggest that a comma `,` is not valid at the start of a property: they can start with backtick \`, underscore `_` or letter.
But it would also imply that a property starting with a comment character `#` is not valid either -- which you just said is valid. So I'm not sure anymore. Can you clarify please?

fmoor:
> I read the [docs on identifiers in EdgeQL](https://github.com/golang/go/blob/986b04c0f12efa1c57293f147a9e734ec71f0363/src/encoding/json/encode.go#L64-L104), which suggest that a comma , is not valid at the start of a property: they can start with backtick `, underscore _ or letter.
But it would also imply that a property starting with a comment character # is not valid either -- which you just said is valid. So I'm not sure anymore. Can you clarify please?

Identifier names are quoted with ``` ` ``` so ```select { `,inline` := 10 };``` is a valid query.

```
edgedb> select { `,inline` := 10 };
{{,inline: 10}}
edgedb> select { `#inline` := 10 };
{{#inline: 10}}
edgedb> select { `$inline` := 10 };
error: EdgeQLSyntaxError: backtick-quoted name cannot start with char `$`
  ┌─ query:1:10
  │
1 │ select { `$inline` := 10 };
  │          ^ error
```

I just learned that there were quoted identifiers my self so this was news to me. No identifier can start with `$` on purpose so that it can be used for special things like this tag. Also, it seems like struct tags are domain specific. json maps to a different set of possible keys or property names than EdgeDB does. I think the `$` prefix is the way to go here.

das7pad:
> I just learned that there were quoted identifiers my self so this was news to me. No identifier can start with `$` on purpose so that it can be used for special things like this tag. Also, it seems like struct tags are domain specific. json maps to a different set of possible keys or property names than EdgeDB does. I think the `$` prefix is the way to go here.

So in the _internal_ representation, identifiers/properties can start with any character other than `@`/ `$` and the _text layer_ needs to quote them as documented. The Golang struct tags would see the internal representation, hence the quoting is not applicable and we need to support a wider charset for the first character. Got it, thanks!

I agree with you that `$inline` is the way to go then. I've amended the code changes and update the PR title/body.

fmoor:
Sorry for the delay here. @das7pad, would you rebase on edgedb/edgedb-go:master (we recently fixed a bug in the CI) and fix the lint error and then I can merge this?

das7pad:
> Sorry for the delay here. @das7pad, would you rebase on edgedb/edgedb-go:master (we recently fixed a bug in the CI) and fix the lint error? Then I can merge this.

Done, @fmoor  :)
The other day I noticed some memory corruption when serializing complex structs with inlined fields that were filled from a db query. I could not yet pin-point the root cause. As a start I wrote a minimal reproducer (added as integration test into `query_test.go`), but it does not seem to trigger it.
Would you mind holding off merging this PR until I am able to find the root cause please?

fmoor:
> Would you mind holding off merging this PR until I am able to find the root cause please?

No problem.

das7pad:
TL;DR this PR is ready to merge :)

I was not able to reproduce any kind of memory corruption in either small test beds for this feature or in bigger benchmarks that exercise this code as a bye-product/have all kinds of complex structs.

I did find one cause for nil-ptr-de-reference (fix in https://github.com/edgedb/edgedb-go/pull/212) that could be the trigger of what I considered a memory corruption. It's too bad that I did not keep stack-traces stored somewhere.
Back then I was recreating postgres/edgedb docker containers every now and then when I hit a wall with db-migrations. It's more than plausible that connection setups were failing on retry when the db was temporarily down.

In other news I finished the migration of all my mongodb/bson usage to edgedb and did not spot any issues with the de-serialization layer for multiple levels of struct nesting.
```
$ grep -r pkg/models/ -e '$inline' | wc -l
152
```
I'm very happy with edgedb so far! Keep it up!

PR #205, opened 2022-03-23
Add NewRetryOptions() function

fmoor:


options.go
```diff
@@ -208,6 +208,11 @@ func (r RetryRule) WithBackoff(fn RetryBackoff) RetryRule {
 	return r
 }
 
+// NewRetryOptions returns the default retry options.
+func NewRetryOptions() RetryOptions {
```
elprans, line 212: Perhaps `NewDefaultRetryOptions`?

fmoor: We already have `NewTxOptions()` and `NewRetryRule()` that return the default value for their respective types.

elprans: So you are supposed to call `New..()` and then mutate properties directly?

fmoor: Sort of. It has a method chaining API.
```go
ruleA := edgedb.NewRetryRule()
ruleB := edgedb.NewRetryRule()
edgedb.NewRetryOptions().WithDefault(ruleA).WithCondition(NetworkError, ruleB)
```

elprans (approved):

PR #210, opened 2022-04-25
Document example embeding Optional in a struct

fmoor:


elprans (approved):

fmoor:
> Would be nice to add some examples to module docs too
added

PR #212, opened 2022-05-12
fix nil-ptr-deref caused by overwriting reconnectingConn.conn with nil

das7pad:
Plenty of places are accessing `reconnectingConn.conn` directly, without copying it first and doing a `nil` check on it. A failing reconnect attempt is currently setting the `.conn` pointer to `nil`, making all these direct access call-sites panic from the nil pointer de-reference.

This PR is storing the new `.conn` value locally and only overwriting `reconnectingConn.conn` on the happy path.


fmoor:
Thanks for this @das7pad

PR #213, opened 2022-05-12
Add bounds checks on query position hint in error message generation

das7pad:
I noticed out-of-bounds access on query lines when generating hints on the error code path.

This PR is adding a few bounds checks.

Unfortunately I did not keep the query for reproducing. Sorry about that.


fmoor:
Thanks for the fix! It would be helpful to have an example query that reproduces this behavior for testing if you are able to recreate it.

PR #217, opened 2022-05-14
Use global math/rand instead of rolling own

das7pad:
It is probably an oversight that there exists a global random source in the `math/rand` package?

Side note: The exported functions in `math/rand` are safe for concurrent access too.


fmoor:
This is not a oversight. From the `math/rand` docs:
 
> Top-level functions, such as Float64 and Int, use a default shared Source that produces a deterministic sequence of values each time a program is run.

To have random numbers that are not deterministic the generator has to be seeded. If edgedb-go seeds the global random generator it may conflict with users who want to seed the global generator in a particular way (or intentionally not seed it). So to avoid this conflict edgedb-go needs a private generator.

das7pad:
Good point. Using the zero-seeded source, two processes that never called a `math/rand` function would get the same jitter sequences when getting disconnected at the same time. I'm going to close this PR then :slightly_smiling_face: .

PR #218, opened 2022-05-22
Cache type ids immediately after reading them

das7pad:
On first glance it looked like this could yield empty `idPair`s. It required a few more reads (including the buff.Reader code) to understand that the loop cannot exit with `r.Err == nil` when the server did not sent all the messages yet -- with one remaining assumption: the server always sends `PrepareComplete` ahead of `ReadyForCommand`.

I think its simpler to just move the cache call right next to the population of the `idPair` :slightly_smiling_face: . WDYT?

_Side note: I've also fixed the typo in `cardinality` and dropped useless copying of already copied UUID bytes._

### DIFF
--- a/granularflow.go
+++ b/granularflow.go
@@ -184,7 +184,6 @@ func (c *protocolConnection) prepare(r *buff.Reader, q *gfQuery) error {
 
 	var (
 		err error
-		ids idPair
 	)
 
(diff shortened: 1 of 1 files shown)


fmoor (approved):
LGTM

PR #221, opened 2022-06-08
Protocol v1.0

fmoor:


granularflow1pX.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+func (c *protocolConnection) optimistic1pX(
+	r *buff.Reader,
+	q *query,
+	cdcs *codecPair,
+) error {
+	// When descriptors are returned the codec ids sent didn't match the
+	// server's.  The codecs should be rebuilt with the new descriptors and the
+	// execution retried.
```
elprans, line 72: This is not correct.  When the output type descriptor is not matched, the client is supposed to receive the correct one and proceed decoding the data messages, no execution retry is necessary.

fmoor: Ok. What about if the input descriptor is wrong? Does that warrant an execution retry?

elprans: For now, no, because it's a much rarer condition and is likely non-recoverable anyway.

fantix (commented):
The API to set aliases/globals/config in the state is a TODO, right?

fmoor:
> The API to set aliases/globals/config in the state is a TODO, right?

added

elprans (approved):

PR #234, opened 2022-07-21
Support range types

fmoor:


internal/edgedbtypes/range.go
```diff
@@ -0,0 +1,1054 @@
...
+package edgedbtypes
+
+import "encoding/json"
+
+type emptyRangeJSON struct {
+	Empty bool `json:"empty"`
+}
+
+// NewRangeInt32 creates a new RangeInt32 value.
+func NewRangeInt32(
```
elprans, line 26: How do I construct an empty range?

fmoor: pass equal upper and lower bounds where at least one of them is not inclusive

fmoor: I was being conservative on the API here. We can add `NewEmptyRangeX` or something if the current API is to obscure.

elprans (approved):

PR #240, opened 2022-11-12
Skip inaccessible files/folders when generating

fmoor:
When we don't have permission to read a file/folder in the project tree skip it instead of failing.


fmoor:
Maybe this should go behind a flag?

nsidnev (approved):

PR #256, opened 2022-12-22
Add cloud-credentials/secret_key parsing

fmoor:
Depends on https://github.com/edgedb/shared-client-testcases/pull/18

### DIFF
--- a/internal/client/connutils.go
+++ b/internal/client/connutils.go
@@ -20,11 +20,15 @@ import (
 	"crypto/sha1"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
@@ -34,10 +38,19 @@ import (
 
 	"github.com/edgedb/edgedb-go/internal/edgedbtypes"
 	"github.com/edgedb/edgedb-go/internal/snc"
+	"github.com/sigurn/crc16"
 )
 
-var isDSNLike = regexp.MustCompile(`(?i)^[a-z]+://`)
-var isIdentifier = regexp.MustCompile(`^[A-Za-z_][A-Za-z_0-9]*$`)
+var (
+	isDSNLike      = regexp.MustCompile(`(?i)^[a-z]+://`)
+	instanceNameRe = regexp.MustCompile(
+		`^([A-Za-z_]\w*)(?:/([A-Za-z_]\w*))?$`,
+	)
+	crcTable       *crc16.Table = crc16.MakeTable(crc16.CRC16_XMODEM)
+	base64Encoding              = base64.StdEncoding.WithPadding(
+		base64.NoPadding,
+	)
+)
 
 type connConfig struct {
 	addr               dialArgs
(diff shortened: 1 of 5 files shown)


internal/client/connutils.go
```diff
@@ -34,10 +38,19 @@ import (
...
 
-var isDSNLike = regexp.MustCompile(`(?i)^[a-z]+://`)
-var isIdentifier = regexp.MustCompile(`^[A-Za-z_][A-Za-z_0-9]*$`)
+var (
+	isDSNLike      = regexp.MustCompile(`(?i)^[a-z]+://`)
+	instanceNameRe = regexp.MustCompile(
+		`^([A-Za-z_]\w*)(?:/([A-Za-z_]\w*))?$`,
+	)
+	crcTable       *crc16.Table = crc16.MakeTable(crc16.CRC16_XMODEM)
+	base64Encoding              = base64.StdEncoding.WithPadding(
```
fantix, line 50:
```suggestion
	base64Encoding              = base64. URLEncoding.WithPadding(
```

I think this is the alternative used in JWT

internal/client/connutils.go
```diff
@@ -1217,3 +1363,128 @@ func findEdgeDBTOML(paths *cfgPaths) (string, error) {
...
+		if err != nil {
+			return fmt.Errorf(errMsg, err)
+		}
+
+		profile := "default"
+		if r.profile.val == nil {
+			p, ok := os.LookupEnv("EDGEDB_CLOUD_PROFILE")
+			if ok {
+				r.setProfile(p, "EDGEDB_CLOUD_PROFILE environment variable")
+			}
```
fantix, line 1411: Sorry this was my bad in a previous version of the edgedb-python PR - `EDGEDB_CLOUD_PROFILE` should be set just as other connection params do (looks like you've already done that), not here.

fantix (approved):
Other than the 2 comments, this looks good to me 👍

internal/client/connutils.go
fmoor: Hm... If this is not supposed to be here maybe [this test case](https://github.com/edgedb/shared-client-testcases/blob/main/connection_testcases.json#L2784) needs to be revised? Without this `LookupEnv()` call environment variables are only read if there is no host or port or instance or dsn or credentials file name or credentials json blob. However the test case sited seems to expect the environment to be read even though there is an instance set. Maybe I'm misunderstanding something.

fantix: You're totally right, I did something like [this](https://github.com/edgedb/edgedb-python/pull/405/commits/c4383ecc07cb03805ce94f1cc28f555aca365dbd#diff-74188b7ca9fbcd019d3646664133b00e705e3453eb08dfc1726dde21a1e29f3bR527-R585) in edgedb-python to pass this test - get env var of the profile in the options layer, so that it's both functional on the options layer and the env var layer, and thus overwrites deeper layers like project stashed `cloud-profile`. I don't know maybe we should just allow setting cloud profile in the API options too (and fix this test to expect a failure) 🤷‍♂️ so that we have no exceptions like this.

fmoor: Seems like this type of special case can make the API confusing to use.

fantix: Right, let's fix it then?

fmoor: Sure, I'll change this test to be a failure for now. Then we can add the secretKey option later.

fmoor: https://github.com/edgedb/shared-client-testcases/pull/22

Issue #245, opened 2022-12-08
edgeql-go generates private only code


joeyak:
When edgeql-go is run, the generated function names and return values are lowercase, which makes them inaccessible.

I have `select_enabled_auth.edgeql`
```
SELECT Auth {
    id,
    access_token,
    refresh_token,
    token_expires
}
FILTER .user.channel_settings.enabled = true;
```

and a `queries.go` file to generate the code and export the queries
```go
//go:generate edgeql-go

package database

var (
	SelectEnabledAuth = selectEnabledAuth
)
```

the generated code is

```go
// Code generated by github.com/edgedb/edgedb-go/cmd/edgeql-go DO NOT EDIT.

package database

import (
	"context"
	_ "embed"
	"time"

	"github.com/edgedb/edgedb-go"
)

//go:embed select_enabled_auth.edgeql
var selectEnabledAuthCmd string

// selectEnabledAuth
// runs the query found in
// select_enabled_auth.edgeql
func selectEnabledAuth(
	ctx context.Context,
	client *edgedb.Client,
) ([]struct {
	id            edgedb.UUID `edgedb:"id"`
	access_token  string      `edgedb:"access_token"`
	refresh_token string      `edgedb:"refresh_token"`
	token_expires time.Time   `edgedb:"token_expires"`
}, error) {
	var result []struct {
		id            edgedb.UUID `edgedb:"id"`
		access_token  string      `edgedb:"access_token"`
		refresh_token string      `edgedb:"refresh_token"`
		token_expires time.Time   `edgedb:"token_expires"`
	}

	err := client.Query(
		ctx,
		selectEnabledAuthCmd,
		&result,
	)

	return result, err
}

// selectEnabledAuthJSON
// runs the query found in
// select_enabled_auth.edgeql
// returning the results as json encoded bytes
func selectEnabledAuthJSON(
	ctx context.Context,
	client *edgedb.Client,
) ([]byte, error) {
	var result []byte

	err := client.QueryJSON(
		ctx,
		selectEnabledAuthCmd,
		&result,
	)
	if err != nil {
		return nil, err
	}

	return result, nil
}
```
When I go to use the returned values, I cannot call on the fields. The generated files would work fine if it's in the  same package, but I have the database layer exported out.
![image](https://user-images.githubusercontent.com/27922090/206471986-640a05f3-96b2-4a75-8323-765aee61653e.png)

**Expected behavior**
[...]

PR #262, opened 2023-02-14
Add -mixedcaps option to edgeql-go

fmoor:
The new `-mixedcaps` option changes snake_case names in shapes to MixedCaps names in go structs.

```
select User {
    first_name,
    last_name,
}
```

Becomes:
```go
type MyQueryResult struct {
    FirstName string `edgedb:"first_name"`
    LastName  string `edgedb:"last_name"`
}
```

Without this option enabled names are used exactly as they appear in shapes. For example the query above would produce the following struct:
```go
type MyQueryResult struct {
    first_name string `edgedb:"first_name"`
    last_name  string `edgedb:"last_name"`
}
```

working on https://github.com/edgedb/edgedb-go/issues/245


sebastiean:
Can this be reviewed? It would be an extremely useful feature

fantix (approved):
Sorry for the delay! I think this is good.

Issue #277, opened 2023-07-08
Add support for `database` config in project metadata


elprans:
See example implementation in [edgedb-python](https://github.com/edgedb/edgedb-python/pull/442)

PR #278, opened 2023-07-11
Implement database config in project dir

mgaidia:
Implements https://github.com/edgedb/edgedb-go/issues/277


fmoor (changes requested):
This looks good to me. The [shared-client-testcases](https://github.com/edgedb/shared-client-testcases) git submodule should be updated too. I think there are some tests added there for this use case.

fmoor (approved):
Thanks!

PR #280, opened 2023-07-24
Move rst docs gen from website repo

jaclarke:


.github/workflows/push-docs.sh
```diff
@@ -0,0 +1,11 @@
+git config user.name edgedb-ci
+git config user.email releases@edgedb.com
+git add .
+
+if ! git diff --cached --exit-code > /dev/null;
```
fmoor, line 5: What do you think about requiring the PR author to run `make gendocs` and adding this check to the regular workflow as a lint? This way the rendered rst changes are part of the code review. It is somewhat redundant in terms of reviewing documentation content but could help catch errors in the rst rendering code particularly as godoc syntax evolves.

fmoor (approved):

PR #285, opened 2023-09-26
Document the $inline tag

fmoor:
fixes https://github.com/edgedb/edgedb-go/issues/284

### DIFF
--- a/rstdocs/index.rst
+++ b/rstdocs/index.rst
@@ -148,7 +148,7 @@ optional.
 
 .. code-block:: go
 
-    type User Struct {
+    type User struct {
         edgedb.Optional
         Email string `edgedb:"email"`
     }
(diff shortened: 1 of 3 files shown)


rstdocs/index.rst
```diff
@@ -174,6 +174,21 @@ using sets as parameters.
 Nested structures are also not directly allowed but you can use `json <https://www.edgedb.com/docs/edgeql/insert#bulk-inserts>`_
 instead.
 
+By default EdgeDB will ignore embedded structs when marshaling/unmarshaling.
+To treat an embedded struct's fields as part of the parent struct's fields
+tag the embedded struct with `edgedb:$inline`.
```
raddevon, line 179:
```suggestion
tag the embedded struct with ``edgedb:$inline``.
```

rstdocs/index.rst
```diff
@@ -174,6 +174,21 @@ using sets as parameters.
 Nested structures are also not directly allowed but you can use `json <https://www.edgedb.com/docs/edgeql/insert#bulk-inserts>`_
 instead.
 
+By default EdgeDB will ignore embedded structs when marshaling/unmarshaling.
+To treat an embedded struct's fields as part of the parent struct's fields
```
raddevon, line 178: Make sure I've interpreted this correctly. I wasn't 100% certain how to interpret it.
```suggestion
To treat an embedded struct's fields as part of the parent struct's fields,
```

raddevon (approved):
Two minor change suggestions, but it looks good otherwise!

rstdocs/index.rst
fmoor: Good catch!

The `rst` files are generated from the go documentation comments. Go documentation comments don't have anything analogous to the double back ticks in `rst`. The back ticks here are literally what the user should put in their source `` `edgedb:"$inline"` ``. I guess the generator needs to be updated to escape the back tics.

Also looks like I forgot the quotes!

PR #290, opened 2023-10-25
Add multi range support

diksipav:
Update descriptors.
Add multirange codec.
Add simple multi range test.

**NOTE**: I'll add more and better tests.

### DIFF
--- a/internal/client/types_test.go
+++ b/internal/client/types_test.go
@@ -7846,7 +7846,7 @@ func TestSendAndReceiveInt32MultiRange(t *testing.T) {
 		false,
 	)
 
-	query := "SELECT <array<range<int32>>>$0"
+	query := "SELECT <multirange<int32>$0"
 	err := client.QuerySingle(ctx, query, &result, multiRange)
 	require.NoError(t, err)
 	assert.Equal(t, multiRange, result)
--- a/internal/client/types_test.go
+++ b/internal/client/types_test.go
@@ -7869,21 +7869,6 @@ func TestEmptyMultiRange(t *testing.T) {
 	assert.Equal(t, emptyMultiRange, result)
 }
 
-func TestMultiRangeContains(t *testing.T) {
-	if !serverHasMultiRange(t) {
-		t.Skip("server lacks std::MultiRange support")
-	}
-
-	ctx := context.Background()
-
-	var result bool
-
-	query := "select contains(multirange([range(1, 5), range(8,10)]), 9)"
-	err := client.QuerySingle(ctx, query, &result)
-	require.NoError(t, err)
-	assert.Equal(t, true, result)
-}
-
 func TestCustomSequenceTypeHandling(t *testing.T) {
 	ddl := `
 		CREATE SCALAR TYPE SampleSequence extending std::sequence;
--- a/internal/client/types_test.go
+++ b/internal/client/types_test.go
@@ -6701,7 +6701,7 @@ func TestReceiveOptionalArray(t *testing.T) {
 	})
 }
 
-func TestSendOptioanlArray(t *testing.T) {
+func TestSendOptionalArray(t *testing.T) {
 	ctx := context.Background()
 	var result struct {
 		Val []int64 `edgedb:"val"`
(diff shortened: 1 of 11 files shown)


internal/codecs/codecs.go
```diff
@@ -102,6 +102,8 @@ func BuildEncoder(
 		return buildArrayEncoder(desc, version)
 	case descriptor.Range:
 		return buildRangeEncoder(desc, version)
+	case descriptor.MultiRange:
+		return buildMultiRangeEncoder(desc, version)
```
nsidnev, line 106: looks like the same thing needs to be done in `BuildEncoderV2` below and also you need to add lines to build decoders in `BuildDecoder` and `BuildDecoderV2` in the same file

internal/descriptor/descriptor_v2.go
```diff
@@ -140,6 +140,18 @@ func PopV2(
...
+			name := r.PopString()
+			r.PopUint8() // schema_defined
+			ancestors := scalarFields2pX(r, descriptorsV2, false)
+			fields := []*FieldV2{{
+				Desc: descriptorsV2[r.PopUint16()],
+			}}
+			err := assertArrayDimensions(r)
+			if err != nil {
+				return V2{}, err
+			}
```
nsidnev, line 153: I'm not sure that we need that here and for v1. If I got it right, multiranges descriptor looks similar to this

```c
struct MultiRangeTypeDescriptor {
    // Indicates that this is an
    // MultiRange Type descriptor.
    uint8   tag = 12;

    // Schema type ID.
    uuid    id;

    // Schema type name.
    string  name;

    // Whether the type is defined in the schema
    // or is ephemeral.
    bool    schema_defined;

    // Number of ancestor scalar types.
    uint16  ancestors_count;

    // Indexes of ancestor scalar type descriptors
    // in ancestor resolution order (C3).
    uint16  ancestors[ancestors_count];

    // Range type descriptor index.
    uint16  type;
};
```

`assertArrayDimensions` will attempt to read below the descriptor bytes and possibly corrupt the remainder of the `describe` message.

internal/client/types_test.go
```diff
@@ -7806,3 +7806,48 @@ func TestSendAndReceiveRangeLocalDate(t *testing.T) {
...
+	)
+
+	multiRange[1] = types.NewRangeInt32(
+		types.NewOptionalInt32(1),
+		types.NewOptionalInt32(10),
+		true,
+		false,
+	)
+
+	query := "SELECT <array<range<int32>>>$0"
```
nsidnev, line 7849: this will just test an array of ranges, not what we need. the query probably should be more like this:
```
select <multirange<int32>>$0
```

this will force the EdgeDB to require and return multiranges instead of arrays

diksipav: Yees! I should test multirange construct

internal/client/types_test.go
```diff
@@ -7807,6 +7807,83 @@ func TestSendAndReceiveRangeLocalDate(t *testing.T) {
...
+		true,
+		false,
+	)
+
+	multiRange[1] = types.NewRangeInt32(
+		types.NewOptionalInt32(8),
+		types.NewOptionalInt32(10),
+		true,
+		false,
+	)
```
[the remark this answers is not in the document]
diksipav: Hey yes, I can, I will update... I used somewhere in tests aliases and somewhere not. Kinda wanted to get a feeling how things work, but its probably cleaner to keep to one style.

fmoor (commented):
Looks good overall  :+1:

fmoor:
@diksipav how is this going? Is this waiting on a review?

internal/client/types_test.go
```diff
@@ -7807,6 +7807,83 @@ func TestSendAndReceiveRangeLocalDate(t *testing.T) {
...
+func TestMultiRangeContains(t *testing.T) {
+	if !serverHasMultiRange(t) {
+		t.Skip("server lacks std::MultiRange support")
+	}
+
+	ctx := context.Background()
+
+	var result bool
+
+	query := "select contains(multirange([range(1, 5), range(8,10)]), 9)"
```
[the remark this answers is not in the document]
diksipav: Hm, my idea was here to test that `multirange([range(1, 5), range(8,10)]) contains 9`.

fmoor (approved):

PR #302, opened 2024-03-12
Expose a common interface between Client and Tx

lsdch:
This PR introduces the `Executor` interface which exposes all querying functions that are common to both `Client` and `Tx`.
Example use-case:

```go
func CreateItem(db edgedb.Executor, name string) (item Item, err error) {
  query := `select (insert Item { name := $0 }) { name };`
  err = db.QuerySingle(context.Background(), query, &item, name)
  return 
}

// Using Client
client, _ := edgedb.CreateClient(ctx, edgedb.Options{})
CreateItem(client, "Item name")

// Using Tx
client.Tx(context.Background(), func(ctx context.Context, tx *edgedb.Tx) error {
  CreateItem(tx, "Item name")
  // Transaction may then be rolled back, e.g. if this is a unit-test
  return errors.New("Rollback")
})
```

### DIFF
--- a/export.go
+++ b/export.go
@@ -71,6 +71,10 @@ type (
 	// ErrorTag is the argument type to Error.HasTag().
 	ErrorTag = edgedb.ErrorTag
 
+	// Executor is a common interface between Client and Tx,
+	// that can run queries on an EdgeDB database.
+	Executor = edgedb.Executor
+
 	// IsolationLevel documentation can be found here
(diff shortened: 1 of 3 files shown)


fmoor (approved):
Looks good. Thanks!

lsdch:
Thanks! Linter seems to be happy now :slightly_smiling_face:

fmoor:
This change was released in v0.16.0.

PR #304, opened 2024-03-18
Add convenience functions to transform between edgedb.Duration and time.Duration

lsdch:
This is meant to improve usability of `Duration` values and better integrate with the Go ecosystem.
This also makes the unit (micro/nanoseconds) difference  between `Duration` and `time.Duration` more obvious, hopefully reducing the likelihood of introducing hard to detect bugs such as: 
```go
func ConvertDuration(d edgedb.Duration) time.Duration {
  return time.Duration(int64(d)) 
  // change from microseconds to nanoseconds representation is not taken care of, 
  // and the compiler can not help to prevent this
}

### DIFF
--- a/internal/edgedbtypes/datetime.go
+++ b/internal/edgedbtypes/datetime.go
@@ -660,6 +660,26 @@ func (d Duration) String() string {
 	return strings.Join(buf, "")
 }
 
+// AsNanoseconds returns [time.Duration] represented as nanoseconds,
+// after transforming from Duration microsecond representation.
+// Returns an error if the Duration is too long and would cause an overflow of
+// the internal int64 representation.
+func (d Duration) AsNanoseconds() (time.Duration, error) {
+	if int64(d) > math.MaxInt64/int64(time.Microsecond) ||
+		int64(d) < math.MinInt64/int64(time.Microsecond) {
+		return time.Duration(0), fmt.Errorf(
+			"Duration is too large to be represented as nanoseconds",
+		)
+	}
+	return time.Duration(d) * time.Microsecond, nil
+}
+
+// DurationFromNanoseconds creates a Duration represented as microseconds
+// from a [time.Duration] represented as nanoseconds.
+func DurationFromNanoseconds(d time.Duration) Duration {
+	return Duration(math.RoundToEven(float64(d) / 1e3))
+}
+
 // NewOptionalDuration is a convenience function for creating an
 // OptionalDuration with its value set to v.
 func NewOptionalDuration(v Duration) OptionalDuration {
--- a/internal/edgedbtypes/datetime_test.go
+++ b/internal/edgedbtypes/datetime_test.go
@@ -19,6 +19,7 @@ package edgedbtypes
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: internal/edgedbtypes/datetime.go [hunk 1]
```diff
@@ -660,6 +660,25 @@ func (d Duration) String() string {
 	return strings.Join(buf, "")
 }
 
+// AsNanoseconds returns [time.Duration] represented as nanoseconds,
+// after transforming from Duration microsecond representation.
+// Returns an error if the Duration is too long and would cause an overflow of
+// the internal int64 representation.
+func (d Duration) AsNanoseconds() (time.Duration, error) {
+	if int64(d) > math.MaxInt64/int64(time.Microsecond) {
```
fmoor, line 668: Does this account for negative duration? Would you add some tests for large and small negative durations?

fmoor (changes requested):
I like it!

internal/edgedbtypes/datetime.go
lsdch: Ah yes, I forgot to account for negative duration !

internal/edgedbtypes/datetime_test.go
```diff
@@ -758,6 +759,32 @@ func TestUnmarshalDuration(t *testing.T) {
...
+		{Duration(math.MaxInt64 / 1000), false},
+		{Duration(0), false},
+		{Duration(math.MinInt64), true},
+		{Duration(math.MinInt64 / 100), true},
+		{Duration(math.MinInt64 / 1000), false},
+	}
+
+	for _, c := range cases {
+		t.Run(c.input.String(), func(t *testing.T) {
+			_, err := c.input.AsNanoseconds()
```
fmoor, line 778: It would be nice if this also asserted the result value.

internal/edgedbtypes/datetime.go
```diff
@@ -660,6 +660,26 @@ func (d Duration) String() string {
...
+			"Duration is too large to be represented as nanoseconds",
+		)
+	}
+	return time.Duration(d) * time.Microsecond, nil
+}
+
+// FromNanoseconds creates a Duration represented as microseconds
+// from a [time.Duration] represented as nanoseconds.
+func FromNanoseconds(d time.Duration) Duration {
+	return Duration(d.Microseconds())
```
fmoor, line 680: I was rereading the documentation and rediscovered that clients are supposed to [round to the nearest even](https://www.edgedb.com/docs/reference/bindings/datetime#precision).

There is a reference implementation [here](https://github.com/edgedb/edgedb-go/blob/0be826dd40630c314c470cc0c2b35fccb4b787de/internal/codecs/datetime.go#L91).

internal/edgedbtypes/datetime_test.go
```diff
@@ -758,6 +759,32 @@ func TestUnmarshalDuration(t *testing.T) {
...
 }
 
+func TestNanosecondsDuration(t *testing.T) {
+	cases := []struct {
+		input    Duration
+		mustFail bool
+	}{
+		{Duration(math.MaxInt64), true},
+		{Duration(math.MaxInt64 / 100), true},
+		{Duration(math.MaxInt64 / 1000), false},
```
fmoor, line 769: Would you add a test case for the max and min `std::duration` values from EdgeDB as well as max+1 and min-1.

fmoor (changes requested):
Thanks for getting to this so quickly! Its been a while so I'm remembering a few more details as I look at this again.

internal/edgedbtypes/datetime.go
lsdch: Is using `math.RoundToEvent` acceptable ? I find it more straightforward
```go
func FromNanoseconds(d time.Duration) Duration {
	return Duration(math.RoundToEven(float64(d) / 1e3))
}
```

internal/edgedbtypes/datetime_test.go
lsdch: I used min/max values from [tests of the Rust implementation](https://github.com/edgedb/edgedb-rust/blob/master/edgedb-protocol/tests/datetime_chrono.rs) after [this recommendation](https://www.edgedb.com/docs/reference/bindings/datetime#precision). Please let me know if I was mistaken :slightly_smiling_face:

internal/edgedbtypes/datetime.go
fmoor: Yea! I had no idea that existed :exploding_head:

internal/edgedbtypes/datetime_test.go
fmoor: Sorry I got turned around. My intention was to have a test for the largest and smallest possible values for `AsNanoseconds()` and then one unit out side the valid range on both ends. The idea being that if someone changes this function in the future they won't accidentally shrink the window or change the error behavior.

So its not the min/max edgedb.Duration values as the input. It is the min/max time.Duration values as the output.

lsdch: Makes sense ! I added cases on both sides of each boundary, plus some cases for arbitrary values within range

internal/edgedbtypes/datetime.go
```diff
@@ -660,6 +660,26 @@ func (d Duration) String() string {
...
+		return time.Duration(0), fmt.Errorf(
+			"Duration is too large to be represented as nanoseconds",
+		)
+	}
+	return time.Duration(d) * time.Microsecond, nil
+}
+
+// FromNanoseconds creates a Duration represented as microseconds
+// from a [time.Duration] represented as nanoseconds.
+func FromNanoseconds(d time.Duration) Duration {
```
fmoor, line 679: I think this needs to be added to [export.go](https://github.com/edgedb/edgedb-go/blob/master/export.go) to be publicly accessible.

fmoor (approved):
Looks good. One last detail I forgot about.

internal/edgedbtypes/datetime.go
lsdch: Thanks! I also changed the name for clarity

fmoor:
Any last changes before I merge?

fmoor:
released in `v0.17.0`

PR #312, opened 2024-05-05
Add a Value() function to optional types

chirino:
The Value() is like the Get() function except that it returns the  zero value when the option is not set.  This can make the optimal type easier to use in some scenarios.

### DIFF
--- a/internal/edgedbtypes/bool.go
+++ b/internal/edgedbtypes/bool.go
@@ -38,6 +38,9 @@ type OptionalBool struct {
 // Get returns the value and a boolean indicating if the value is present.
 func (o OptionalBool) Get() (bool, bool) { return o.val, o.isSet }
 
+// Value returns the value or the zero value if not set.
+func (o OptionalBool) Value() bool { return o.val }
+
 // Set sets the value.
 func (o *OptionalBool) Set(val bool) {
 	o.val = val
--- a/internal/edgedbtypes/bytes.go
+++ b/internal/edgedbtypes/bytes.go
@@ -36,6 +36,9 @@ type OptionalBytes struct {
 // Get returns the value and a boolean indicating if the value is present.
 func (o OptionalBytes) Get() ([]byte, bool) { return o.val, o.isSet }
 
+// Value returns the value or the zero value if not set.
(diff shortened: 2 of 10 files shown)


fmoor:
Thanks for the PR!

> This can make the optimal type easier to use in some scenarios.

Can you elaborate on this?

We chose the `Get()` API instead of using pointers because we wanted maximal type safety. Adding this `Value()` method seems to largely circumvent the type safety provided by `Get()`.

chirino:
This is just idomatic go.  Go does not have optional types.  In many cases, when the optional is not set, the caller wants to get the nil type.

consider this usecase: I'm writing a http handler to expose some data form EdgeDB as an API for my front end and I need to covert the edgeDB result struct to struct that will be sent as a JSON result.  Using the value function this is really easy: 

```go
//  in gin request hander, we do a call to edgeDB to lookup a user by id: 
	record, _ := dbGet(c, client, id)

	type User struct {
		Id       edgedb.UUID `json:"id"`
		FullName string      `json:"full_name"`
		Picture  string      `json:"picture,omitempty"` // zero values are special!
	}

	c.JSON(http.StatusOK, User{
		Id:       record.id,
		FullName: record.fullName,
		Picture:  record.picture.Value(), // I just want the zero value when not set.
	})
```
Without the Value() function, the caller has to deal with more toil:

```
        picture, _ := record.picture.Value()
	c.JSON(http.StatusOK, User{
		Id:       record.id,
		FullName: record.fullName,
		Picture:  picture,
	})
```

It's quite common across the go ecosystem to use the zero values to mean something like the value not being set.

fmoor:
I understand where you are coming from. While it is common for the zero value to mean unset it is not a rule. I don't think we are ready to commit to this kind of change quite yet. It is easy enough to write a convenience function for this.

```go
type Getter[T any] interface {
	Get() (T, bool)
}

func Value[T any](thing Getter[T]) T {
	v, _ := thing.Get()
	return v
}
```

chirino:
What's the downside to adding the Value() function?  It seems like having the method can make life easier for many of your users without much harm to anyone else.   You guys do want to attract go developers to use your project right?

PR #313, opened 2024-05-05
Fix #287: Adds -pubtypes, -pubfuncs, and -jsontag CLI flags

### LINKED ISSUE #287: Add mechanism for generating public query functions and types
Currently `edgeql-go` generates private query functions and return types. The user can export these types but it requires a few lines of code for each query. There should be a way to make the generator generate public instead of private functions and types.

Maybe this should be configurable on a per query basis?

chirino:
* -pubtypes make the generated types public
* -pubfuncs makes the generated functions public
* -jsontag add a `json:”...”` field tag

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 edgedb
 *.test
 /.vscode
+/.idea
--- a/cmd/edgeql-go/gen.go
+++ b/cmd/edgeql-go/gen.go
@@ -28,7 +28,7 @@ func generateType(
 	desc descriptor.Descriptor,
 	required bool,
 	path []string,
-	mixedCaps bool,
+	opts options,
 ) ([]goType, []string, error) {
 	var (
 		err     error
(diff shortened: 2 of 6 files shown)


fmoor:
Awesome, thanks for this!

Would you add test cases for this? Copy the `cmd/edgeql-go/testdata/no-args` directory and rename once for each of the new options. Adjust the directory content for the behavior with the option being tested. Then add a test case for each new directory [here](https://github.com/edgedb/edgedb-go/blob/dabd2d1f752b59397335c94723895241ed10fc58/cmd/edgeql-go/endtoend_test.go#L38).

It would be great if the new options were added to the [usage example](https://github.com/edgedb/edgedb-go/blob/dabd2d1f752b59397335c94723895241ed10fc58/cmd/edgeql-go/doc.go#L33) in the docs.

I'm wondering if the `-jsontag` should perhaps be more general. Maybe it could be `-tags` and accept name+strategy pairs like `json:snakecase`. This would allow adding multiple tags and being able to choose the name transformation strategy. I think it would be nice to split this into its own PR. I don't want the design discussion for this to hold up the other two options.

chirino:
Feel free to make any changes you need.  I thought I would send you this contribution to get you started.  But I'm fine using a fork.

fmoor:
Closing in favor of https://github.com/edgedb/edgedb-go/pull/317 and https://github.com/edgedb/edgedb-go/pull/318.

PR #326, opened 2024-11-27
Add gel as a valid dsn scheme.

dnwpark:


internal/client/connutils.go
```diff
@@ -1209,9 +1209,9 @@ func parseDSN(dsn string) (*url.URL, map[string]string, error) {
 		return nil, nil, fmt.Errorf("could not parse DSN %q: %w", dsn, err)
 	}
 
-	if uri.Scheme != "edgedb" {
+	if uri.Scheme != "edgedb" && uri.Scheme != "gel" {
 		return nil, nil, fmt.Errorf(
-			`scheme is expected to be "edgedb", got %q`, uri.Scheme)
+			`scheme is expected to be "edgedb" or "gel", got %q`, uri.Scheme)
```
fmoor, line 1214: I think this can just say `"gel"` and drop `"edgedb"`. We'll still accept `"edgedb"`, but we want people to move away from it so don't mention it as an option.

fmoor (commented):
Looks good! Just a minor nit, and fix the failing test and this can be merged.

fmoor:
Hm... I also forgot that this also requires a change in the shared-client-testcases. That change is already there so you can update that submodule if you like. Unfortunately this will require that you make all the tests pass and there are a few other new gel related things in the shared-client-testcases changes.

PR #328, opened 2024-12-18
Add support for protocol v3 and add `QuerySQL` and `ExecuteSQL` client methods

jaclarke:
See JS implementation: https://github.com/edgedb/edgedb-js/pull/1131, and python: https://github.com/edgedb/edgedb-python/pull/534/files

### DIFF
--- a/internal/client/client_test.go
+++ b/internal/client/client_test.go
@@ -191,6 +191,62 @@ func TestQuerySingleJSONMissingResult(t *testing.T) {
 		"the \"out\" argument must be *[]byte or *OptionalBytes, got *string")
 }
 
+func TestQuerySQL(t *testing.T) {
+	ctx := context.Background()
+
+	var version int64
+	err := client.QuerySingle(ctx, "SELECT sys::get_version().major", &version)
+	assert.NoError(t, err)
+
+	if version >= 6 {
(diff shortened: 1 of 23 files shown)


internal/client/client_test.go
```diff
@@ -191,6 +191,64 @@ func TestQuerySingleJSONMissingResult(t *testing.T) {
...
+func TestQuerySQL(t *testing.T) {
+	ctx := context.Background()
+
+	var version int64
+	err := client.QuerySingle(ctx, "SELECT sys::get_version().major", &version)
+	assert.NoError(t, err)
+
+	if version >= 6 {
+		// err = client.ExecuteSQL(ctx, "select 1")
+		// assert.NoError(t, err)
```
fmoor, line 203: Should this be removed?

fmoor (approved):
LGTM

internal/client/client_test.go
jaclarke: No, good catch. I commented it out because it was failing, with the intention to go back and fix before I pushed the branch, but it seems I forgot. It looks like it's failing due to a server bug: https://github.com/edgedb/edgedb/issues/8185, where the server is returning an unexpected type descriptor for executeSQL queries.

PR #337, opened 2025-02-24
Improve documentation by removing assignment exports

fmoor:
Almost the entire client API was defined in `internal/` and then made public using assignment like this `type Client = internal.Client`. This made the documentation annoying to navigate. This change moves the public API definition into public packages so that the documentation is more straight forward.

The diff here is huge. It will probably make more sense to view the rendered docs pages than to look at the diff. Check out this branch and run the following in the project root directory.

```bash
go run golang.org/x/pkgsite/cmd/pkgsite@latest
```

Compare this to https://pkg.go.dev/github.com/edgedb/edgedb-go


zackelan (approved):
this diff is too large to truly review, but I really like the improvement in the docs output

nutgood (approved):
Looked at the docs, not the code, but this looks great.

PR #354, opened 2025-03-07
Use geldata/setup-go instead of edgedb/setup-edgedb

fmoor:


.github/workflows/tests.yml
```diff
@@ -60,15 +60,15 @@ jobs:
...
-          wsl-shell-user: edgedb
+          wsl-shell-user: gel
           additional-packages:
             ca-certificates
             curl
 
-      - name: Install EdgeDB
-        uses: edgedb/setup-edgedb@v1
+      - name: Install Gel
+        uses: geldata/setup-gel@358410cc5618fa956d16f1b7157a8c86f9b517cf # v1.4
```
elprans, line 69: Since this is an action we control and trust, it's fine to keep using `v1` and avoid the hassle of bumping hashes as we update it.

PR #358, opened 2025-03-10
Add examples for client query methods

fmoor:
To view the rendered documentation run the following in a local checkout of this branch:
```bash
go run golang.org/x/pkgsite/cmd/pkgsite@latest
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,17 +20,24 @@ import (
 	"context"
 
 	"github.com/geldata/gel-go/gelcfg"
+	"github.com/geldata/gel-go/gelerr"
 	"github.com/geldata/gel-go/geltypes"
 	gel "github.com/geldata/gel-go/internal/client"
 )
 
 // CreateClient returns a new client. The client connects lazily. Call
-// Client.EnsureConnected() to force a connection.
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -20,17 +20,24 @@ import (
...
 	"github.com/geldata/gel-go/geltypes"
 	gel "github.com/geldata/gel-go/internal/client"
 )
 
 // CreateClient returns a new client. The client connects lazily. Call
-// Client.EnsureConnected() to force a connection.
+// [Client.EnsureConnected] to force a connection.
+//
+// We recommend using an empty [gelcfg.Options] struct.
+// [Read more] about the recommended ways to configure client connections.
```
scotttrinh, line 32: Not a blocker, but I think maybe we can provide some more of the reasoning here before pointing them off to the docs?

Maybe something like:

```suggestion
// Instead of providing connection details directly, we recommend connecting
// using projects locally, and environment variables for remote instances,
// providing an empty [gelcfg.Options] struct here.
// [Read more] about the recommended ways to configure client connections.
```

Or something like that?

scotttrinh (approved):
Looks good! One small completely optional suggestion.

PR #384, opened 2025-07-16
Add an example webserver using the gel-go client

gscho:
While working on learning gel and the `gel-go` SDK I decided to create a simple web server to test my knowledge. I've added the example code to the repo in case others may also find it useful when learning to use `gel-go`.


scotttrinh:
Amazing, and thanks for the contribution! I think we want example apps like this to live in https://github.com/geldata/gel-examples but I'll see if @fmoor wants to keep this one here as a decent "Hello world" kind of example? If he wants to move them into the examples repo, we can add a section to the README here that points to the go examples to make it easier to discover.

fmoor:
I'm open to either place.

